PR #7, opened 2019-09-06
Added self-hosted cluster setup guide

micahhausler:
*Issue #, if available:*
N/A
*Description of changes:*

Added self-hosted k8s setup guide

### DIFF
--- a/SELF_HOSTED_SETUP.md
+++ b/SELF_HOSTED_SETUP.md
@@ -1,0 +1,145 @@
+# Self-hosted Kubernetes setup
+
+If you are running your own Kubernetes cluster, there are several steps required for this feature to work.
+
+This feature requires Kubernetes 1.12 or greater.
+
+## Projected Token Signing Keypair
+
+The first thing required is a new key pair for signing and verifying projected
+service account tokens. This can be done using the following `ssh-keygen`
+commands.
+
+```bash
+# Generate the keypair
+PRIV_KEY="sa-signer.key"
+PUB_KEY="sa-signer.key.pub"
+PKCS_KEY="sa-signer-pkcs8.pub"
+# Generate a key pair
+ssh-keygen -t rsa -b 2048 -f $PRIV_KEY -m pem
+# convert the SSH pubkey to PKCS8
+ssh-keygen -e -m PKCS8 -f $PUB_KEY > $PKCS_KEY
+```
+
+## Public Issuer
+
+As of 1.16, Kubernetes does not include an OIDC discovery endpoint itself (see
+[kubernetes/community#1190](https://github.com/kubernetes/enhancements/pull/1190)),
+so you will need to put your public signing key somewhere that AWS STS can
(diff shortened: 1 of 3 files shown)


SELF_HOSTED_SETUP.md
```diff
@@ -0,0 +1,127 @@
...
+ssh-keygen -e -m PKCS8 -f $PUB_KEY > $PKCS_KEY
+```
+
+## Public Issuer
+
+As of 1.16, Kubernetes does not include an OIDC discovery endpoint itself (see
+[kubernetes/community#1190](https://github.com/kubernetes/enhancements/pull/1190)),
+so you will need to put your public signing key somewhere that AWS STS can
+discover it. This example, we will create one in a public S3 bucket, but you
+could host the following documents any way you'd like on a different domain.
```
jqmichael: any where?

micahhausler: Eh, either or. I was probably thinking “any way” like dynamically instead of a static JSON document

SELF_HOSTED_SETUP.md
```diff
@@ -0,0 +1,127 @@
...
+# to the $PUB_KEY file from the beginning
+--service-account-key-file
+
+# Path to the signing (private) key ($PRIV_KEY)
+--service-account-signing-key-file
+
+# The default audience for tokens that do not specify an audience. Tokens with
+this audience can be used to authenticate with Kubernetes
+# `--api-audiences` is for v1.13+, `--service-account-api-audiences` in v1.12
+--api-audiences
```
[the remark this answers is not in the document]
micahhausler: I'll clarify this in the guide.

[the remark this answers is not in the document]
micahhausler: I added a section below on audiences and expanded the comment above

SELF_HOSTED_SETUP.md
```diff
@@ -0,0 +1,143 @@
...
+
+# The issuer URL, or "https://$ISSUER_HOSTPATH" from above.
+--service-account-issuer
+```
+
+## Audiences
+
+The above `--api-audiences` flag sets an `aud` value for tokens that do not
+request an audience, and the API server requires that any projected tokens used
+to for pod to API server authentication must have this audience set.
```
anguslees:
```suggestion
for pod to API server authentication must have this audience set.
```

SELF_HOSTED_SETUP.md
```diff
@@ -0,0 +1,143 @@
...
+export ISSUER_HOSTPATH=$HOSTNAME/$S3_BUCKET
+```
+
+### Create the OIDC discovery and keys documents
+
+Part of the OIDC spec is to host an OIDC discovery and a keys JSON document.
+Lets create these:
+
+```bash
+# Get the sha of the pubilc key and use it as the key id
```
nckturner: public

PR #9, opened 2019-09-11
Removed key id from jwks JSON hack script 

micahhausler:
*Issue #, if available:*

N/A

*Description of changes:*
Kubernetes doesn't set the key ID field until 1.16. Once 1.16 lands, this will need to be changed

CC @siwyd

### DIFF
--- a/SELF_HOSTED_SETUP.md
+++ b/SELF_HOSTED_SETUP.md
@@ -54,8 +54,6 @@ Part of the OIDC spec is to host an OIDC discovery and a keys JSON document.
 Lets create these:
 
 ```bash
-# Get the sha of the public key and use it as the key id
-KID=$(sha1sum $PUB_KEY | awk '{print $1}')
 cat <<EOF > discovery.json
 {
     "issuer": "https://$ISSUER_HOSTPATH/",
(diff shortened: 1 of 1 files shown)


micahhausler:
@mogren Yes, K8s just never set a `kid` until just two weeks ago in kubernetes/kubernetes#78502.

STS treats the empty string `""` as the kid value, and requires tokens have the same kid `""`.

siwyd:
@micahhausler What will happen when an 1.15 cluster is upgraded to 1.16? Newly launched pods will then suddenly have the kid in the token? I imagine those would fail to then authenticate with AWS as the kid will no longer match if the JWKS only has one key with kid = "". I just tried a JWKS with two keys, the same key repeated, but one has kid = "" and one has kid = "foobar". I can still authenticate as the kid = "" seems to be selected properly by AWS OIDC. Perhaps it would be better for now to generate a JWKS that has the same key repeated, one with kid = "" and one with kid having the correct hash. That way things work on 1.15 and before and should keep on working with 1.16 instead of suddenly breaking? Or am I misinterpreting something and things will not break when 1.16 is released?

Issue #17, opened 2019-11-08
Jsonpatch operation fails if automountServiceAccountToken is false and no volumes are defined


FaHeymann:
**What happened**:

A ServiceAccount with `automountServiceAccountToken: false` and the EKS IAM Annotation failed to spawn a Pod from a Job definition that did not define Volumes on the Pod with the following Error

```
Events:
  Type     Reason        Age                From            Message
  ----     ------        ----               ----            -------
  Warning  FailedCreate  16s (x2 over 26s)  job-controller  Error creating: Internal error occurred: Internal error occurred: jsonpatch add operation does not apply: doc is missing path: "/spec/volumes/0"
```

https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L158-L177 probably needs to check if `/spec/volumes` already exists. If `automountServiceAccountToken` on the ServiceAccount is true, there is always a volume present on the Pod, if set to false this is no longer the case though.

**How to reproduce it (as minimally and precisely as possible)**:

```yaml
apiVersion: v1
automountServiceAccountToken: false
kind: ServiceAccount
metadata:
  name: test
  namespace: test
  annotations:
    eks.amazonaws.com/role-arn: <redacted>
---
apiVersion: batch/v1
kind: Job
metadata:
  name: test
  namespace: test
spec:
  template:
    spec:
      restartPolicy: "Never"
      containers:
      - image: nginx
        name: my-container
      securityContext:
        fsGroup: 1000
      serviceAccountName: test
```

**Anything else we need to know?**:

**Environment**:
- AWS Region: eu-central-1
- EKS Platform version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.1
- Kubernetes version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.version`): 1.14
- Webhook Version:

tarvitz:
Sorry, for breaching in, my message is related to `Error creating: Internal error occurred: Internal error occurred: jsonpatch add operation does not apply: doc is missing path` error message, whenever it appears. Googling the source of the problem leads to this issue 😉 .

The "**root**" problem is that json-patch message can not be applied to the resource (i.e. it's malformed). So basically, any who develops mutating webhook might get such the confusing error. So if you have it => try to observe the object you try to json-patch, this will lead you to the source of the problem.

PR #18, opened 2019-11-30
Fix jsonpatch operation if no volumes are present

FaHeymann:
Fixes https://github.com/aws/amazon-eks-pod-identity-webhook/issues/17

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -154,33 +154,49 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string) []p
 		containers = append(containers, container)
 	}
 
-	patch := []patchOperation{
-		patchOperation{
-			Op:   "add",
-			Path: "/spec/volumes/0",
-			Value: corev1.Volume{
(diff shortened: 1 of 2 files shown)


jqmichael:
Nice fix!

thomaschaaf:
@jaypipes this is still causing problems for us. In which release can this bugfix be found?

dumain18:
This is also causing problems for us. 
When adding `automountServiceAccountToken: false` to our service account we see the following error when trying to apply a deployment: 

```
  - lastTransitionTime: "2020-09-14T12:21:48Z"
    lastUpdateTime: "2020-09-14T12:21:48Z"
    message: 'Internal error occurred: jsonpatch add operation does not apply: doc
      is missing path: "/spec/volumes/0"'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
```

We see this error when trying to apply a pod:

`Error from server (InternalError): error when creating "testpod.yaml": Internal error occurred: jsonpatch add operation does not apply: doc is missing path: "/spec/volumes/0"
`

Issue #20, opened 2019-12-06
Remove the role designation from a Service Account


jasonrichardsmith:
**What would you like to be added**:

I am wondering why a particular role is being associated with the service account.  It looks like we can remove that part and let the containers in the pods set their own role they want to assume using the shared SA JWT token in the pod.  
You can allow the same service account to assume several roles.  I ran into this issue when trying to add service accounts to the sidecar injector.  If a developer sets a SA on a pod for something like access to a particular CRD, I have to annotate the SA or provide a new one.  Either way I have to elevate the side car injector privileges within the cluster.  If the user has their own SA for IAM setup, I have to overwrite that annotation or somehow force a new App Mesh policy into the existing role.

It does not appear that the SA annotation needs to be directly linked to a role.  It just needs to signify this SA token should load a JWT for AWS.

**Why is this needed**:
Several roles in AWS can be associated with a Service Account just like you can bind several K8s roles to a service account.

micahhausler:
Thanks for filing this issue. At this point we cannot change the existing behavior of looking up the annotation on a ServiceAccount because that would result in a breaking change for our customers.

The primary reason we put the annotation on the ServiceAccount rather than on the pod is because the identity within the Kubernetes cluster is what the IAM role's trust policy has specified, not the pod itself. The other reason we lookup the role annotation on the service account is to encourage a least privilege model where one Kubernetes identity assumes a single role. You are correct that a Service Account is capable of assuming multiple roles, however it is more complex to keep track of which roles are bound to a particular SA.

If there is a need for specifying a different role than the annotation, I do want to support that use case. Currently the [webhook short-circuits](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L103-L106) if either `AWS_ROLE_ARN` or `AWS_WEB_IDENTITY_TOKEN_FILE` are present and doesn't modify the container. What do you think about keeping the existing behavior, but if `AWS_ROLE_ARN` is present in a container, not overwriting it with whatever the ServiceAccount specifies? This would have the effect of letting you use different identities for applications using the existing flow, but specifying a different role for a sidecar for something like AppMesh. Would that suit your use case?

jasonrichardsmith:
Thanks for responding.
I just want to follow with a few points.

I agree the trust policy should trust the service account but it is a one to many relationship.  Basically this current implementation ties authz to authn, in theory if not in actual practice.

[...]

PR #22, opened 2019-12-09
Remove roles from Service Account

jasonrichardsmith:
*Issue #, if available:*
#20 

*Description of changes:*
Remove role annotation check from Service Account and load a JWT for any pod that has a role ARN in any env variable.


jasonrichardsmith:
@jaypipes yes philosophical difference on the implementation.  I will close.

PR #26, opened 2019-12-30
Ensure AWS_WEB_IDENTITY_TOKEN_FILE points to correct path on windows pod

niranjan94:
**Issue #:** AWS Support Case ID `6708626561`

**Description of changes:**
If the pod is a windows container (identified by the `nodeSelector` `beta.kubernetes.io/os`), the path to the identity file is modified to be a valid windows path.

**Example:** `C:\var\run\secrets\eks.amazonaws.com\serviceaccount\token` is used instead of `/var/run/secrets/eks.amazonaws.com/serviceaccount/token`

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
+	"strings"
 
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache"
 	"k8s.io/api/admission/v1beta1"
@@ -94,7 +95,7 @@ type patchOperation struct {
 	Value interface{} `json:"value,omitempty"`
 }
 
-func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volName, roleName string) {
+func addEnvToContainer(container *corev1.Container, mountPath, tokenFilePath, volName, roleName string) {
 	reservedKeys := map[string]string{
 		"AWS_ROLE_ARN":                "",
 		"AWS_WEB_IDENTITY_TOKEN_FILE": "",
@@ -118,9 +119,10 @@ func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volNam
 		Name:  "AWS_ROLE_ARN",
 		Value: roleName,
 	})
+
 	env = append(env, corev1.EnvVar{
 		Name:  "AWS_WEB_IDENTITY_TOKEN_FILE",
-		Value: filepath.Join(mountPath, tokenName),
+		Value: tokenFilePath,
 	})
 	container.Env = env
 	container.VolumeMounts = append(
@@ -141,16 +143,27 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string) []p
 		}
 	}
 
+	tokenFilePath := filepath.Join(m.MountPath, m.tokenName)
+
+	betaNodeSelector, _ := pod.Spec.NodeSelector["beta.kubernetes.io/os"]
+	nodeSelector, _ := pod.Spec.NodeSelector["kubernetes.io/os"]
+	if (betaNodeSelector == "windows") || nodeSelector == "windows" {
+		// Convert the unix file path to a windows file path
+		// Eg. /var/run/secrets/eks.amazonaws.com/serviceaccount/token to
+		//     C:\var\run\secrets\eks.amazonaws.com\serviceaccount\token
+		tokenFilePath = "C:" + strings.Replace(tokenFilePath, `/`, `\`, -1)
+	}
+
 	var initContainers = []corev1.Container{}
 	for i := range pod.Spec.InitContainers {
 		container := pod.Spec.InitContainers[i]
-		addEnvToContainer(&container, m.MountPath, m.tokenName, m.volName, roleName)
+		addEnvToContainer(&container, m.MountPath, tokenFilePath, m.volName, roleName)
 		initContainers = append(initContainers, container)
 	}
 	var containers = []corev1.Container{}
 	for i := range pod.Spec.Containers {
 		container := pod.Spec.Containers[i]
-		addEnvToContainer(&container, m.MountPath, m.tokenName, m.volName, roleName)
+		addEnvToContainer(&container, m.MountPath, tokenFilePath, m.volName, roleName)
 		containers = append(containers, container)
 	}
 
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -48,6 +48,54 @@ var rawPodWithoutVolume = []byte(`
 }
 `)
 
+var rawWindowsBetaPodWithoutVolume = []byte(`
+{
+  "apiVersion": "v1",
+  "kind": "Pod",
+  "metadata": {
+	"name": "balajilovesoreos",
+	"uid": "be8695c4-4ad0-4038-8786-c508853aa255"
+  },
+  "spec": {
+	"containers": [
+	  {
+		"image": "amazonlinux",
+		"name": "balajilovesoreos"
+	  }
+	],
+	"serviceAccountName": "default",
+	"nodeSelector": {
+		"beta.kubernetes.io/arch": "amd64",
+		"beta.kubernetes.io/os": "windows"
+	}
+  }
+}
+`)
+
+var rawWindowsPodWithoutVolume = []byte(`
+{
+  "apiVersion": "v1",
+  "kind": "Pod",
+  "metadata": {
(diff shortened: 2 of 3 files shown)


pkg/handler/handler.go
```diff
@@ -118,9 +120,22 @@ func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volNam
...
 	})
+
+	hostTokenFilePath := filepath.Join(mountPath, tokenName)
+	if isWindowsContainer {
+		// Convert the unix file path to a windows file path
+		// Eg. /var/run/secrets/eks.amazonaws.com/serviceaccount/token to
+		//     C:\var\run\secrets\eks.amazonaws.com\serviceaccount\token
+		if absolutePath, err := filepath.Abs(hostTokenFilePath); err != nil {
+			hostTokenFilePath = absolutePath
+		}
```
mogren: If `filepath.Abs` returns an error, would the rest even work? Or would it try to refer to `"C:\tokenName"`?

pkg/handler/handler.go
```diff
@@ -118,9 +120,22 @@ func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volNam
...
+		hostTokenFilePath = strings.Replace(hostTokenFilePath, `/`, `\`, -1)
+		hostTokenFilePath = "C:" + hostTokenFilePath
```
mogren: Nit: It's fine to do this on one line:
```go
hostTokenFilePath = "C:" + strings.Replace(hostTokenFilePath, `/`, `\`, -1)
```

pkg/handler/handler.go
```diff
@@ -141,16 +156,22 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string) []p
 		}
 	}
 
+	isWindowsContainer := false
+	if val, ok := pod.Spec.NodeSelector["beta.kubernetes.io/os"]; ok {
+		isWindowsContainer = val == "windows"
+	}
+
+
```
mogren: Nit: One extra newline.

pkg/handler/handler_test.go
```diff
@@ -73,11 +97,47 @@ var rawPodWithVolume = []byte(`
 }
 `)
 
-func getValidReview(isVolumePresent bool) *v1beta1.AdmissionReview {
-	pod := rawPodWithoutVolume
+var rawWindowsPodWithVolume = []byte(`
```
mogren: Thanks for adding the tests! 👍

pkg/handler/handler.go
niranjan94: I'm not actually very sure as to what cases `filepath.Abs` would return the error. But perhaps we could handle it this way ...


```go
if absolutePath, err := filepath.Abs(hostTokenFilePath); err != nil {
	hostTokenFilePath = absolutePath
} else {
	if !strings.HasPrefix(hostTokenFilePath, "/") {
		hostTokenFilePath = "/" + hostTokenFilePath
	}
}
```

That way we ensure there is a leading slash even if there is an error. But no way to ensure the path is actually correct.

Any suggestions/insights from you on this ?

niranjan94:
Thanks a lot for your review @mogren 😄
Have made the requested changes.

pkg/handler/handler.go
mogren: Well, I think it would probably be better to just error out completely if that fails, since the token file path must be completely wrong at that stage.

niranjan94: Yes. That seems to be a better option. Something like should work

```go
absolutePath, err := filepath.Abs(hostTokenFilePath)
if err != nil {
	klog.Errorf("error resolving absolute path of host token file: %v", err)
	return
}
hostTokenFilePath = "C:" + strings.Replace(absolutePath, `/`, `\`, -1)
```

mogren: I was thinking something even more drastic, like `os.Exit(1)` instead of just a `return`. The token file path should really be correct.

pkg/handler/handler.go
```diff
@@ -94,11 +95,12 @@ type patchOperation struct {
 	Value interface{} `json:"value,omitempty"`
 }
 
-func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volName, roleName string) {
+func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volName, roleName string, isWindowsContainer bool) {
 	reservedKeys := map[string]string{
 		"AWS_ROLE_ARN":                "",
 		"AWS_WEB_IDENTITY_TOKEN_FILE": "",
 	}
+
```
micahhausler: nit: remove extraneous newline

pkg/handler/handler.go
```diff
@@ -94,7 +95,7 @@ type patchOperation struct {
 	Value interface{} `json:"value,omitempty"`
 }
 
-func addEnvToContainer(container *corev1.Container, mountPath, tokenName, volName, roleName string) {
+func addEnvToContainer(container *corev1.Container, mountPath, hostTokenFilePath, volName, roleName string) {
```
niranjan94: @micahhausler we'll need to pass both `mountPath` and `hostTokenFilePath`. as the `mountPath` will only refer to the directory containing the token file and not the file itself. 

Also, I'm not converting `mountPath` to a platform-specific path as kubernetes supports specifying unix-style file paths for windows container workload volume mounts.

Let me know if I should change this to platform-specific path.

pkg/handler/handler.go
```diff
@@ -141,16 +143,29 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string) []p
...
 	}
 
+	hostTokenFilePath := filepath.Join(m.MountPath, m.tokenName)
+	if val, ok := pod.Spec.NodeSelector["beta.kubernetes.io/os"]; ok && val == "windows" {
+		// Convert the unix file path to a windows file path
+		// Eg. /var/run/secrets/eks.amazonaws.com/serviceaccount/token to
+		//     C:\var\run\secrets\eks.amazonaws.com\serviceaccount\token
+		absolutePath, err := filepath.Abs(hostTokenFilePath)
+		if err != nil {
+			klog.Errorf("error resolving absolute path of host token file: %v", err)
```
niranjan94: @micahhausler any suggestions on error handling ? Currently I'm just printing to the log and returning.

pkg/handler/handler.go
```diff
@@ -141,16 +143,29 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string) []p
 		}
 	}
 
+	hostTokenFilePath := filepath.Join(m.MountPath, m.tokenName)
+	if val, ok := pod.Spec.NodeSelector["beta.kubernetes.io/os"]; ok && val == "windows" {
+		// Convert the unix file path to a windows file path
+		// Eg. /var/run/secrets/eks.amazonaws.com/serviceaccount/token to
+		//     C:\var\run\secrets\eks.amazonaws.com\serviceaccount\token
+		absolutePath, err := filepath.Abs(hostTokenFilePath)
```
micahhausler: Why the `Abs` call here? If the `MountPath` wasn't absolute to begin with,  [`filepath.Abs`](https://golang.org/pkg/path/filepath/#Abs) will join with the current working directory _of the webhook server_.

pkg/handler/handler.go
micahhausler: naming nit: can you remove `host`? Its not actually a host path (same for the `updatePodSpec` var name)

pkg/handler/handler.go
micahhausler: If you remove the `Abs` call, you won't need any error handling :)

pkg/handler/handler.go
niranjan94: Makes sense. Let me remove that :)

pkg/handler/handler.go
```diff
@@ -141,16 +143,24 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string) []p
 		}
 	}
 
+	tokenFilePath := filepath.Join(m.MountPath, m.tokenName)
+	if val, ok := pod.Spec.NodeSelector["beta.kubernetes.io/os"]; ok && val == "windows" {
```
micahhausler: Sorry, one last thing. The [documented node OS label for 1.14](https://v1-14.docs.kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels) is `kubernetes.io/os`. Can you have this match either `beta.kubernetes.io/os` or `kubernetes.io/os` and update the documentation? Everything else looks great

niranjan94: @micahhausler done. Added support+tests for both node selectors.

Issue #36, opened 2020-02-17
SELF_HOSTED_SETUP Document Correction


geckofu:
### APIServer Configuration
In [SELF_HOSTED_SETUP.md](https://github.com/geckofu/amazon-eks-pod-identity-webhook/blob/master/SELF_HOSTED_SETUP.md#kubernetes-api-server-configuration) it says you need to add `--service-account-key-file` with the path to the `$PKCS_KEY` file. I have tested with minikube 1.17 and a kops 1.15 cluster, here is the error message:

```
{"log":"Error: invalid authentication config: error reading public key file /var/lib/minikube/certs/sa-signing-pkcs.pub: data does not contain any valid RSA or ECDSA public keys\n","stream":"stderr","time":"2020-02-10T14:34:36.834045998Z"}
```

According to the [apiserver doc](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/), it should be the path to `$PUB_KEY`

Also you should pass the original public key to `--service-account-key-file` so that existing tokens won't fail: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/29

### ControllerManager Configuration
It is missing from the doc that you should also configure controller managers: 

> You must pass a service account private key file to the token controller in the controller-manager by using the --service-account-private-key-file option. The private key will be used to sign generated service account tokens.  [Managing Service Accounts](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#token-controller)

otterley:
@geckofu I think this change (in #37) caused a regression.  The `$PUB_KEY` file is not an PEM-encoded key as required by the apiserver; it is an OpenSSH-format public key.  The `$PKCS_KEY` file is the PEM-encoded key.

As the apiserver documentation states:

> File containing **PEM-encoded** x509 RSA or ECDSA private or public keys, used to verify ServiceAccount tokens.

This issue was raised before in #11 - I fear that we are going back and forth here.

PR #37, opened 2020-02-17
Update SELF_HOSTED_SETUP.md

geckofu:
*Issue #, if available:* 
https://github.com/aws/amazon-eks-pod-identity-webhook/issues/36

*Description of changes:*
As specified in the issue, this pull request updates apiserver and controller-manager configurations in  SELF_HOSTED_SETUP.md

### DIFF
--- a/SELF_HOSTED_SETUP.md
+++ b/SELF_HOSTED_SETUP.md
@@ -131,15 +131,6 @@ In order to use this feature, you'll need to set the following
 --service-account-issuer
 ```
 
-## Controller Manager configuration
-
-You'll need to point controller manager to the new signing private key by setting the following [controller manager flags](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/).
-
-```
-# Path to the signing (private) key ($PRIV_KEY)
---service-account-private-key-file
-```
-
 ## Audiences
 
 The above `--api-audiences` flag sets an `aud` value for tokens that do not
--- a/SELF_HOSTED_SETUP.md
+++ b/SELF_HOSTED_SETUP.md
@@ -108,9 +108,12 @@ In order to use this feature, you'll need to set the following
 [API server flags](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).
 
 ```
-# This flag is likely already specified for legacy service accounts, you can
-# specify this flag multiple times, and you'll need to add this with the path
-# to the $PKCS_KEY file from the beginning
+# Path to the $PUB_KEY file from the beginning. 
+#
+# This flag can be specified for multiple times.
+# There is likely already one specified for legacy service accounts, if not, 
+# it is using the default value. Find out your default value and pass it explicitly
+# (along with this $PUB_KEY), otherwise your existing tokens will fail.
 --service-account-key-file
 
 # Path to the signing (private) key ($PRIV_KEY)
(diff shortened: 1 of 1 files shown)


SELF_HOSTED_SETUP.md
```diff
@@ -128,6 +131,15 @@ In order to use this feature, you'll need to set the following
 --service-account-issuer
 ```
 
+## Controller Manager configuration
```
micahhausler: This is incorrect and needs to be removed. The Controller Manager created keys are stored in etcd and should be separate from the TokenRequest signed keys

geckofu: IIRC, I couldn't get the it working without this controller manager flag. Will test again, thanks for the review, get back to you soon.

geckofu: Yes, I confirmed that pods with new service account tokens can't access apiserver without this controller manager flag point to the new private key file. (Minikube 1.17)

It is used by controller manager to sign service account tokens, and apiserver uses corresponding public key to verify the  token.

```
--service-account-private-key-file string
Filename containing a PEM-encoded private RSA or ECDSA key used to sign service account tokens.

From the controller-manager doc: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/
```

I'm not sure what you mean that controller manager created keys are stored in etcd, could you elaborate more?

SELF_HOSTED_SETUP.md
```diff
@@ -128,6 +131,15 @@ In order to use this feature, you'll need to set the following
...
+
+You'll need to point controller manager to the new signing private key by setting the following [controller manager flags](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/).
```
geckofu: Maybe the phrasing is a bit misleading, I could change it to:

```suggestion
You'll need to the specify the [following flag](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) with the path to the new signing private key file:
```

SELF_HOSTED_SETUP.md
micahhausler: Did you specify `--service-account-key-file` only once on the API server? You'll need to specify it twice: once for the legacy service account verification, and once for the new projected service account token verification.

In regard to key flags:

Old setup
```yaml
kube-apiserver \
  --service-account-key-file=old-key.pub # So the server can validate ServiceAccount tokens signed by the old private key 

kube-controller-manager \
 --service-account-private-key-file=old-key.key # the controller manager creates ServiceAccount tokens that don't expire and are stored in etcd
```

New setup
```yaml
kube-apiserver \
  --service-account-key-file=old-key.pub # So the server can still validate ServiceAccount tokens signed by the old private key 
  --service-account-key-file=new-key.pub # So the server can validate ServiceAccount tokens signed by the new private key 
  --service-account-signing-key-file=new-key.key # So the server can create new projected ServiceAccount tokens

kube-controller-manager \
 --service-account-private-key-file=old-key.key
```

> I'm not sure what you mean that controller manager created keys are stored in etcd, could you elaborate more?

The Controller manager enforces that for every ServiceAccount there are corresponding Secret objects, and those Secrets contain a JSON Web Token (JWT) that does not expire. Secrets (and all other API objects) are stored in etcd.

The _new_ projected service account tokens are created _by the APIserver_ because the API server has the new private key (from `--service-account-signing-key-file`)

geckofu: @micahhausler I have read through the TokenRequest API, you are right, the section about controller manager configuration has been removed.

PR #38, opened 2020-02-27
Inject AWS_DEFAULT_REGION in mutated pods

josselin-c:
*Description of changes:*

Pods using IRSA should have information about what regions they are running in so that they can communicate with the correct AWS endpoints.
Update amazon-eks-pod-identity-webhook to set the AWS_DEFAULT_REGION environment variable into each mutated container if the value isn't set already.

### DIFF
--- a/main.go
+++ b/main.go
@@ -61,6 +61,7 @@ func main() {
 	audience := flag.String("token-audience", "sts.amazonaws.com", "The default audience for tokens. Can be overridden by annotation")
 	mountPath := flag.String("token-mount-path", "/var/run/secrets/eks.amazonaws.com/serviceaccount", "The path to mount tokens")
 	tokenExpiration := flag.Int64("token-expiration", 86400, "The token expiration")
+	region := flag.String("aws-default-region", "", "If set, AWS_DEFAULT_REGION will be set to this value in mutated containers")
 
 	version := flag.Bool("version", false, "Display the version and exit")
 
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -61,6 +61,7 @@ func main() {
 	audience := flag.String("token-audience", "sts.amazonaws.com", "The default audience for tokens. Can be overridden by annotation")
 	mountPath := flag.String("token-mount-path", "/var/run/secrets/eks.amazonaws.com/serviceaccount", "The path to mount tokens")
 	tokenExpiration := flag.Int64("token-expiration", 86400, "The token expiration")
+	region := flag.String("aws-default-region", "", "The region the cluster is running in. We use this value to set AWS_DEFAULT_REGION in mutated containers")
```
micahhausler: nit: Can you remove the first sentence? This is not necessarily the region the _cluster_ is running in, it is the value that is injected into a a container. And also call out that an unset value will not inject anything

pkg/handler/handler_test.go
```diff
@@ -179,28 +179,73 @@ var rawWindowsPodWithVolume = []byte(`
...
+	"name": "balajilovesoreos",
+	"uid": "be8695c4-4ad0-4038-8786-c508853aa255"
+  },
+  "spec": {
+	"containers": [
+	  {
+		"image": "amazonlinux",
+		"name": "balajilovesoreos",
+		"env": [
+		  {"name":"AWS_REGION","value":"paris"}
```
micahhausler: 👍

josselin-c:
Updated doc as requested. Thanks for keeping an eye out for it.

PR #43, opened 2020-03-18
Get region information from EC2Metadata

josselin-c:
Get the region information from the EC2 Metadata API (IMDS) if the
region flag is not set or empty.
Region information is then injected into containers as environment
variables.
With this commit, region information must be passed to the
webhook, either by flag or via the EC2 Metadata API.

### DIFF
--- a/ATTRIBUTIONS.txt
+++ b/ATTRIBUTIONS.txt
@@ -1,0 +1,865 @@
+** github.com/aws/amazon-eks-pod-identity-webhook; version master --
+https://github.com/aws/amazon-eks-pod-identity-webhook
+** github.com/aws/aws-sdk-go; version 1.29.27 --
+https://github.com/aws/aws-sdk-go/
+** github.com/docker/distribution; version v2.7.1 --
+https://github.com/docker/distribution
+** github.com/google/gofuzz; version v1.0.0 -- https://github.com/google/gofuzz
+** github.com/googleapis/gnostic; version v0.2.0 --
+https://github.com/googleapis/gnostic
+** github.com/jmespath/go-jmespath; version master --
+https://github.com/jmespath/go-jmespath
+** github.com/matttproud/golang_protobuf_extensions; version v1.0.1 --
+https://github.com/matttproud/golang_protobuf_extensions
+** github.com/modern-go/concurrent; version master --
+https://github.com/modern-go/concurrent
+** github.com/modern-go/reflect2; version v1.0.1 --
+https://github.com/modern-go/reflect2
+** github.com/opencontainers/go-digest; version v1.0.0 --
+https://github.com/opencontainers/go-digest
+** github.com/prometheus/client_golang; version v0.9.3 --
+https://github.com/prometheus/client_golang
+** github.com/prometheus/client_model; version master --
+https://github.com/prometheus/client_model
+** github.com/prometheus/common; version v0.4.0 --
+https://github.com/prometheus/common
+** github.com/prometheus/procfs; version master --
+https://github.com/prometheus/procfs
+** gopkg.in/yaml.v2; version v2.2.2 -- https://github.com/go-yaml/yaml
+** k8s.io/api; version master -- https://github.com/kubernetes/api
+** k8s.io/apimachinery; version master --
+https://github.com/kubernetes/apimachinery
+** k8s.io/client-go; version v11.0.1 -- https://github.com/kubernetes/client-go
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -100,6 +101,15 @@ func main() {
 	)
 	saCache.Start()
 
+	if *region == "" {
+		ec2MetadataClient := ec2metadatawrapper.New(nil)
+
+		metadataRegion, err := ec2MetadataClient.Region()
+		if err == nil {
+			*region = metadataRegion
+		}
```
jqmichael: Should we fail fast for error?

pkg/ec2metadatawrapper/ec2metadatawrapper.go
```diff
@@ -0,0 +1,64 @@
...
+
+// HttpClient is used to help with testing
+type HttpClient interface {
+	GetInstanceIdentityDocument() (ec2metadata.EC2InstanceIdentityDocument, error)
+	Region() (string, error)
+}
+
+// EC2MetadataClient to used to obtain a subset of information from EC2 IMDS
+type EC2MetadataClient interface {
+	GetInstanceIdentityDocument() (ec2metadata.EC2InstanceIdentityDocument, error)
```
jqmichael: Is this method being used anywhere?

main.go
jaypipes: at a minimum, I think we should print an INFO log message stating that we experienced an error contacting IMDS.

pkg/ec2metadatawrapper/ec2metadatawrapper.go
```diff
@@ -0,0 +1,64 @@
...
+			WithMaxRetries(metadataRetries),
+		))
+		return &ec2MetadataClientImpl{client: ec2metadata.New(awsSession)}
+	}
+	return &ec2MetadataClientImpl{client: client}
+}
+
+// InstanceIdentityDocument returns instance identity documents
+// http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+func (c *ec2MetadataClientImpl) GetInstanceIdentityDocument() (ec2metadata.EC2InstanceIdentityDocument, error) {
```
jaypipes: I don't see this method being called anywhere?

main.go
josselin-c: My thinking was that for someone who runs the webhook on a environment that is not ec2, the call to `ec2MetadataClient.Region()` would fail but the webhook should still continue its execution.
So I guess the underlying question is, is `amazon-eks-pod-identity-webhook` ec2-only?

pkg/ec2metadatawrapper/ec2metadatawrapper.go
josselin-c: No, I used https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ec2metadatawrapper/ec2metadatawrapper.go that's why the method is there.
I will remove that part of the code.

main.go
jqmichael: People run webhook on desktop for testing at least. We provide a flag for case like this. But it is indeed backwards incompatible (users who used to run webhook on desktop successfully have to start passing in a flag). 

But I would argue always injecting the AWS_REGION env variable is the correct behavior in the long term. By not checking the error, we risk not setting that env variable. 

That's the trade-off we need to make (my personal vote is enforcing region to be non-empty and hence fail the webhook if the call to IMDS failed). 

Looking forwards to other's thoughts.

jaypipes: Agree with @jqmichael. Better to error out here.

josselin-c: makes sense, got it.

pkg/ec2metadatawrapper/ec2metadatawrapper_test.go
```diff
@@ -0,0 +1,78 @@
...
+	Version:          "2010-08-31",
+	Region:           "us-east-1",
+	AccountID:        "012345678901",
+	InstanceID:       "i-01234567",
+	BillingProducts:  []string{"bp-01234567"},
+	ImageID:          "ami-12345678",
+	InstanceType:     "t2.micro",
+	PendingTime:      time.Now(),
+	Architecture:     "x86_64",
+}
```
jaypipes: this is fine I suppose but not strictly necessary since the identity doc stuff isn't used

josselin-c: removed

micahhausler (dismissed):
LGTM

PR #47, opened 2020-04-01
Makefile: fix certificate approval

otterley:
*Description of changes:*

Fix error in Makefile that was causing `kubectl certificate approve` to fail due to an extraneous `$` character.


otterley:
You are correct, my bad.  I think the root cause of the issue is in fact a race condition that I mistook for an error in the Makefile.   Withdrawing.

PR #50, opened 2020-04-04
Self-hosting docs: add troubleshooting, prerequisites

otterley:
*Description of changes:*

Some issues came up while trying to get the webhook to work on a cluster that didn't have its
CA properly set up.  These updates may help future users who encounter similar difficulties.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #51, opened 2020-04-06
Improving Error messages and adding comments to Store.go

pshuman-heb:
While attempting to debug the issues we were seeing around #50, we ran into a number of places where the logged output of the webhook sent us down wrong paths to try to fix things. There's an error that is logged, but totally normal and expected on the initial run of the Webhook in a given cluster.

Also, while I was digging in to try to work out what was going on, I did some commenting to try to mentally map out what this code was actually doing that I thought might be helpful to others in the future. Neither of the main functions in this file are named particularly helpfully to the outside observer, so I've tried to provide some context and common usage pointers. 

---

### DIFF
--- a/pkg/cert/store.go
+++ b/pkg/cert/store.go
@@ -45,38 +45,56 @@ func NewSecretCertStore(namespace, secretName string, clientset clientset.Interf
 	}
 }
 
+// Will attempt to read a secret from the Kubernetes Secret store, format and return it.
+// Will error out if the secret isn't set, or the certificate inside is not properly formatted.
 func (s *secretCertStore) Current() (*tls.Certificate, error) {
+	// Try to get a Certificate from Kube Secret.
+	// This will be there if there's been a successful deployment of the webhook before in this cluster.
 	secret, err := s.clientset.CoreV1().Secrets(s.namespace).Get(
 		s.secretName,
 		metav1.GetOptions{},
 	)
+	// Create an appropriate custom Error for if the secret doesn't exist.
 	noKeyErr := certificate.NoCertKeyError(
-		fmt.Sprintf("no cert/key files read at secret %s/%s",
+		fmt.Sprintf("Error: No data found for cert/key files at secret %s/%s",
 			s.namespace,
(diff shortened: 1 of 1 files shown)


pkg/cert/store.go
```diff
@@ -45,38 +45,56 @@ func NewSecretCertStore(namespace, secretName string, clientset clientset.Interf
...
+	// Create an appropriate custom Error for if the secret doesn't exist.
 	noKeyErr := certificate.NoCertKeyError(
-		fmt.Sprintf("no cert/key files read at secret %s/%s",
+		fmt.Sprintf("Error: No data found for cert/key files at secret %s/%s",
 			s.namespace,
 			s.secretName))
+	// If we had an error getting the secret, exit.
 	if err != nil {
 		klog.Errorf("Error fetching secret: %v", err.Error())
+		klog.Error("This is expected if this is the initial deploy in this cluster.")
```
wongma7, line 65: these might not appear after each other in case logs get printed async by various things , I would merge it into one message

pkg/cert/store.go
```diff
@@ -45,38 +45,56 @@ func NewSecretCertStore(namespace, secretName string, clientset clientset.Interf
...
+	// Try to get a Certificate from Kube Secret.
+	// This will be there if there's been a successful deployment of the webhook before in this cluster.
 	secret, err := s.clientset.CoreV1().Secrets(s.namespace).Get(
 		s.secretName,
 		metav1.GetOptions{},
 	)
+	// Create an appropriate custom Error for if the secret doesn't exist.
 	noKeyErr := certificate.NoCertKeyError(
-		fmt.Sprintf("no cert/key files read at secret %s/%s",
+		fmt.Sprintf("Error: No data found for cert/key files at secret %s/%s",
```
wongma7, line 59: personally I would make it consistent to NOT prefix each error with "Error", that should be evident from the log level (klog.Error). But, I dont really care that much so long as it's consistent.

wongma7 (approved):

Issue #61, opened 2020-06-02
re-invoking webhook exits early


yangkev:
**What happened**:
If the webhook is re-invoked (per [reinvocationPolicy: IfNeeded](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy)), the 2nd invocation exits without injecting any of the environment variables / volume mounts.

E.g., say you have mutating webhook `foo-webhook` which adds some sidecar containers, deployed along-side the `pod-identity-webhook` in a cluster.

Consider the following admission webhook execution scenario when admitting a `Pod` with a container, `containerA`.

1. `pod-identity-webhook` runs, modifies `containerA`
2. `foo-webhook` runs, adds `containerB`, `containerC`
3. `pod-identity-webhook` runs, and [does nothing, since a `volume` field is already present](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/4286d35c28724afa8059857a3fd9387e5c8d61ce/pkg/handler/handler.go#L177-L182)

Now `containerB` and `containerC` do not get environment variables / volume mounts added.

**What you expected to happen**:
I expect on re-invoking the `pod-identity-webhook` that it will continue to inject the service account token volume mount and environment variables on new containers.

**How to reproduce it (as minimally and precisely as possible)**:
This test case for [`TestEnvUpdate`](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/4286d35c28724afa8059857a3fd9387e5c8d61ce/pkg/handler/handler_test.go#L411-L457)
```
{
  "apiVersion": "v1",
  "kind": "Pod",
  "metadata": {
	"name": "balajilovesoreos",
	"uid": "be8695c4-4ad0-4038-8786-c508853aa255"
  },
  "spec": {
	"containers": [
	  {
		"image": "amazonlinux",
		"name": "balajilovesoreos",
	  }
	],
	"serviceAccountName": "default",
	"volumes": [
	  {
	    "name": "aws-iam-token"
	  }
	]
  }
}
```

Expected Patch:
```
[
    {
        "op": "add",
        "path": "/spec/containers",
        "value": [
            {
                "env": [
                    {
                        "name": "AWS_ROLE_ARN",
                        "value": "arn:aws:iam::111122223333:role/s3-reader"
                    },
                    {
                        "name": "AWS_WEB_IDENTITY_TOKEN_FILE",
[...]

PR #62, opened 2020-06-03
patch pod spec even if it's already been patched

yangkev:
Before, the updatePodSpec function would return immediately if it detected that the
aws-iam-token volume was already present.

This patch modifies the behavior so that the function continues to
inject environment variables and volume mounts as needed.

Specifically, the behavior of injecting env vars, volume mounts, and
volumes is now:

1. For each env var in "AWS_ROLE_ARN, AWS_WEB_IDENTITY_TOKEN_FILE,
AWS_REGION, AWS_DEFAULT_REGION", inject if not already present in
container spec

2. Inject volume mount if not already present in container spec

3. Inject volume if not already present in pod spec

Fixes https://github.com/aws/amazon-eks-pod-identity-webhook/issues/61

### DIFF
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -159,6 +159,31 @@ var rawPodWithIAMTokenVolume = []byte(`
 	"containers": [
 	  {
 		"image": "amazonlinux",
+		"name": "balajilovesoreos"
+	  }
+	],
+	"serviceAccountName": "default",
+	"volumes": [
+	  {
+	    "name": "aws-iam-token"
+	  }
+	]
+  }
+}
+`)
+
+var rawPodWithIAMTokenVolumeAndVolumeMount = []byte(`
+{
+  "apiVersion": "v1",
+  "kind": "Pod",
+  "metadata": {
+	"name": "balajilovesoreos",
+	"uid": "be8695c4-4ad0-4038-8786-c508853aa255"
+  },
+  "spec": {
+	"containers": [
+	  {
+		"image": "amazonlinux",
 		"name": "balajilovesoreos",
 		"env": [
 			{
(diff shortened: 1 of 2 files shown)


yangkev:
@micahhausler I think folks who want to adopt pod-identity-webhook and use the sidecar pattern will find this useful. Could someone take a look?

pkg/handler/handler_test.go
```diff
@@ -121,6 +121,48 @@ var rawPodWithVolume = []byte(`
 }
 `)
 
+var rawPodWithIAMTokenVolume = []byte(`
```
josselin-c, line 124: Could you make this test more representative of what this patch is doing?
The idea is that  `rawPodWithIAMTokenVolume` should have a `volume` declared but `env` and `volumeMounts` should not be present. 

`validPatchIfIAMTokenVolumePresent` would then prove that with this patch we are adding `env` and `volumeMounts` correctly now.

josselin-c (changes requested):
I think the test needs to be improved a bit to better test what this patch is doing. Otherwise looks good to me.

pkg/handler/handler_test.go
yangkev: Thanks for the review! I think the main behavior we want to test is that`volumeMounts` and `env` are added once, _and only once_, regardless of if the `volume` is already present.

`rawPodWithIAMTokenVolume` (as it stands) tests the case where 1) `volume` present, 2) `volumeMount` present, no duplicate `volumeMount` added.

I can see how this can be mis-leading, so I've separated the cases for whether a `volumeMount` is present or not.

yangkev (commented):
Also added a test that `initContainer`'s get modified as well

josselin-c (approved):
LGTM, thank you!

luciano-nbs:
Hi guys, i think i'm hitting this issue in our EKS cluster, how do we know if this fix is in our EKS cluster? I couldn't find any information on releases. thanks

PR #79, opened 2020-09-22
deploy: add sideEffects to webhook

### LINKED ISSUE #46: admission webhook "iam-for-pods.amazonaws.com" does not support dry run
**What happened**:
Executing `kubectl diff` against a deployment using IAM roles for service accounts (IRSA) resulted in the error "Error from server (BadRequest): admission webhook "iam-for-pods.amazonaws.com" does not support dry run".

It would be nice to support this feature so it is part of our workflow.

**What you expected to happen**:
Expected to see the diff
**How to reproduce it (as minimally and precisely as possible)**:
kubectl diff targeting an EKS cluster.
**Anything else we need to know?**:

**Environment**:
- AWS Region: ap-southeast-2
- EKS Platform version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.1
- Kubernetes version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.version`): 1.15
- Webhook Version:

sjenning:
fixes #46

Adds `sideEffects: None` to the `MutatingWebhookConfiguration` definition so that dry run operations will succeed.

### DIFF
--- a/deploy/mutatingwebhook.yaml
+++ b/deploy/mutatingwebhook.yaml
@@ -6,6 +6,7 @@ metadata:
 webhooks:
 - name: pod-identity-webhook.amazonaws.com
   failurePolicy: Ignore
+  sideEffects: None
   clientConfig:
     service:
       name: pod-identity-webhook


jqmichael (approved):

jqmichael:
Thanks!

PR #86, opened 2020-10-20
Support `fs-group` annotation to add PodSecurityPolicies for non-root containers

NickMeves:
*Issue #, if available:*
N/A

*Description of changes:*

A barrier to entry to using service account token volume projection is its `root` ownership with `0600` permissions by default (not customizable natively until Kubernetes v1.19?).

This leave two options, each with downsides:
- Run the containers as `root` (not great from a security perspective)
- Add a `securityContext/fsGroup` to the pod spec to grant a non-root container user access to the projected token (This is potentially difficult when integrating with upstream code or Helm charts).

This PR adds the ability to specify an `eks.amazonaws.com/fs-group` annotation on either the service account or the pod. The mutating webhook will add a `securityContext/fsGroup` in addition to the volume, mount & container env updates.
- If `securityContext/fsGroup` is already configured on a pod, this does nothing. Existing fsGroup GID on the pod spec takes precedence
- If the annotation is present on both the pod & its service account, the pod's annotation takes precedence

I've added unit tests in addition to manually testing all the configuration scenarios in a local development environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,12 @@ This webhook is for mutating pods that will require AWS IAM access.
       annotations:
         eks.amazonaws.com/role-arn: "arn:aws:iam::111122223333:role/s3-reader"
     ```
-4. All new pod pods launched using this Service Account will be modified to use
+4. Optionally, add the `eks.amazonaws.com/fs-group` annotation to the service
+   account or to the pod's annotations with an integer GID. This will add a
+   `PodSecurityPolicy` with that GID as the `fsGroup` to applicable pods. This
+   is necessary for non-root containers to access the projected service account
+   token which will be owned by root with 0600 permissions.
+5. All new pod pods launched using this Service Account will be modified to use
    IAM for pods. Below is an example pod spec with the environment variables and
    volume fields added by the webhook.
     ```yaml
--- a/main.go
+++ b/main.go
@@ -101,6 +101,7 @@ func main() {
 	saCache.Start()
 
 	mod := handler.NewModifier(
+		handler.WithAnnotationPrefix(*annotationPrefix),
 		handler.WithExpiration(*tokenExpiration),
 		handler.WithMountPath(*mountPath),
 		handler.WithServiceAccountCache(saCache),
@@ -123,7 +124,6 @@ func main() {
 	metricsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "ok")
 	})
-
 
 	tlsConfig := &tls.Config{}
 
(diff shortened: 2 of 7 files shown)


josselin-c:
Hello @NickMeves, thanks for the PR.
I am a bit reluctant to add an AWS specific knob to do something that can already be done using standard Kubernetes constructs.

If a pod needs to access account tokens and the chart has support for running the container as non-root but does not support setting a `securityContext/fsGroup`, meaning the pod won't work in this case, then I think the issue lays in that chart.

You say 

> This is potentially difficult when integrating with upstream code or Helm charts

Could you elaborate a bit more, also do you have specific Charts that exhibit that issue?

NickMeves:
I just did a quick spot check -- it looks like a lot of the vendor supported have at least had this for the last year or so (I did a spot check on Elastic and Mongo). So my perception might be dated.

My concern is more in the 3rd party/individual charts that spring up to support pure Open Source projected. The trend I've seen in general is the following:

- Most supported `podAnnotations`, so things like `kube2iam` worked fine.
- After you all hit the scene, I saw a much larger uptick in charts starting to support injecting custom existing service accounts or allowing service account annotations.
- I've seen slower adoption on `securityContext` (and some even adding `PodSecurityPolicy` support in the charts) -- mostly by more mature providers like I mentioned above.

But I'm having trouble finding a good example, it looks like even 3rd party individual ones I'm spot checking have started to add `securityContext` support to `deployments` or `statefulsets` in the last couple of months.

NickMeves:
No worries if you want to keep this out -- we can maintain a fork.

I just wanted to offer it up in case you found it useful. In our organization's initial evaluation of IRSA, we almost discounted it due to the perception containers had to run as root.

So maybe if not this optional annotation to help, more explicit notes in the `securityContext/fsGroup` non-root solution in the main README?

NickMeves:
The other area beyond Helm is custom app deployment pipelines that might use something like buildpacks & creating Kubernetes resources directly. We maintain our own, so it was easy enough for me to add in the `securityContext/fsGroup` to `deployments` as needed for any apps that wanted IRSA.

This annotation addition would've let me manage it all with the mutating webhook instead of volumes, mounts & envvars with the webhook & the fsGroup separately in our own code.

I haven't really done an inventory of what the vendor space for app deployment on Kubernetes supports (E.g. something like Hashicorp's new Waypoint https://www.hashicorp.com/blog/announcing-waypoint)

josselin-c:
Thanks @NickMeves for all the additional details (didn't know about Waypoint!). I think we can close this for now as support for setting `fsGroup` should come from the deployment tool and there are no clear cases where this support can't be added when missing.

Following your advice, I opened #88  in order to document how users can configure their containers to run as non-root.

PR #87, opened 2020-10-21
Add support for external TLS renewal

stromnet:
After following the guide on https://www.openshift.com/blog/fine-grained-iam-roles-for-openshift-applications to setup AWS integration in our OpenShift cluster, we noticed after a few weeks that the newly created CSRs where not automatically approved. Not sure if that is done on other k8s distributions, but not on OC (4.5). 
An alternative solution of handling the TLS certificates is to use what OC calls "Service serving certificates", i.e. let OC generate & renew certificates, and provide them via a secret. Some more details on that particular problem here, but not really relevant for the PR: https://github.com/sabre1041/openshift-aws-iam-webhook-integration/issues/3

This change adds support for the `--external-tls-renewal` flag which will make the k8s Secret "readonly", i.e only read but never update it. Instead of having the k8s go-client `certificate_manager.Manager` renew it via CSR API, it will just try to re-read it from the secret when it is is nearing expiration time.

The actual implementation is pretty much a copy of the `certificate_manager.Manager` impl, but this allows delegation to a arbitrary method to "Load" it. Unfortunately the original impl does not permit re-use, so had to copy a bit..
As for reading the secret, it uses the existing SecretCertStore impl.

In addition, this also adds reload support for when using external file rather than secrets (`--in-cluster=false`). Actually started that way, with the secret mounted on a volume, before I decided to just read it from the secret from the code. And with the abstraction it was quite straightforward anyway.

### DIFF
--- a/main.go
+++ b/main.go
@@ -54,7 +54,8 @@ func main() {
 	inCluster := flag.Bool("in-cluster", true, "Use in-cluster authentication and certificate request API")
 	serviceName := flag.String("service-name", "pod-identity-webhook", "(in-cluster) The service name fronting this webhook")
 	namespaceName := flag.String("namespace", "eks", "(in-cluster) The namespace name this webhook and the tls secret resides in")
-	tlsSecret := flag.String("tls-secret", "pod-identity-webhook", "(in-cluster) The secret name for storing the TLS serving cert")
+	tlsSecret := flag.String("tls-secret", "pod-identity-webhook", "(in-cluster) The secret name for storing/reading the TLS serving cert")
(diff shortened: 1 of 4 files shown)


texmachina:
@josselin-c any movement or updates on this one? We are running into this as well on OpenShift.

danopia:
To tie some threads together as an outsider, i believe #134 addresses the reload-from-disk aspect of this request. Since kubernetes secrets can be presented as files on disk that should be enough.

stromnet:
Sounds like that yes, if I understand that change correct. pod mounts the external secret as volume and that new code ensures it is re-read automatically. Assuming the fsnotify.Watcher impl actually triggers on such "filesystem" changes, but since this usecase is what that ticket deals with I assume that also have been tested

Issue #83, opened 2020-10-16
Injection sometimes fails on pods created from deployments


kragniz:
**What happened**:

Volume and environment config was not injected into pods created from a deployment.

**What you expected to happen**:

The config to be injected, or some debug information about why the injection didn't happen.

**How to reproduce it (as minimally and precisely as possible)**:

Creating a couple of pods with a direct pod and a deployment:

```
(๑•ᴗ•)⊃━~/g/irsa-reproducer━☆ﾟ cat reproducer.yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: test-app
spec:
  serviceAccount: test-app
  serviceAccountName: test-app
  containers:
  - name: hello
    image: busybox
    command: ['sh', '-c', 'sleep 3600']
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: test-app-token-njbrl
      readOnly: true
  volumes:
  - name: test-app-token-njbrl
    secret:
      defaultMode: 420
      secretName: test-app-token-njbrl
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deploy
  namespace: test-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test-deploy
  template:
    metadata:
      labels:
        app: test-deploy
    spec:
      serviceAccount: test-app
      serviceAccountName: test-app
      containers:
      - name: hello
        image: busybox
        command: ['sh', '-c', 'sleep 3600']
        volumeMounts:
        - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
          name: test-app-token-njbrl
          readOnly: true
      volumes:
      - name: test-app-token-njbrl
        secret:
          defaultMode: 420
          secretName: test-app-token-njbrl
```

And applying together:

```
(๑•ᴗ•)⊃━~/g/irsa-reproducer━☆ﾟ kubectl apply -f reproducer.yaml
```

Results in the directly created pod receiving the injected config, but sometimes not the pod created by the deployment:

```
(๑•ᴗ•)⊃━~/g/irsa-reproducer━☆ﾟ kubectl get pods -n test-app -o jsonpath='{range .items[*]}{@.metadata.name}{" (serviceAccount: "}{@.spec.serviceAccount}{") "}{@.spec.volumes[*].name}{"\n"}{end}'
[...]

PR #90, opened 2020-11-14
Add a debug handler to list cache contents and log mutation decision

shravan-achar:
*Issue #, if available:*

*Description of changes:*

Added a debug endpoint to debug application state. Similar to the kubelet flag --enable-debugging-handlers.

Pre-1.16, audit logs do not have mutation decision annotated. If appropriate logging level is enabled, mutation decision logs could be of help for those clusters.

Issues such as https://github.com/aws/amazon-eks-pod-identity-webhook/issues/83 would be easier to debug

Open to suggestions/feedback on this.

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache"
+	cachedebug "github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache/debug"
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/cert"
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/handler"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -65,6 +66,8 @@ func main() {
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
 
 	version := flag.Bool("version", false, "Display the version and exit")
+
+	debug := flag.Bool("enable-debugging-handlers", false, "Enable debugging handlers. Currently /debug/alpha/cache is supported")
 
 	klog.InitFlags(goflag.CommandLine)
 	// Add klog CommandLine flags to pflag CommandLine
@@ -127,6 +130,16 @@ func main() {
 	metricsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "ok")
 	})
+
+	// Register debug endpoint only if flag is enabled
+	if *debug {
+		debugger := cachedebug.Dumper{
+			Cache: saCache,
+		}
+		// Reuse metrics port to avoid exposing a new port
+		metricsMux.HandleFunc("/debug/alpha/cache", debugger.Handle)
+		// Expose other debug paths
+	}
 
 	tlsConfig := &tls.Config{}
 
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -16,6 +16,7 @@
 package cache
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync"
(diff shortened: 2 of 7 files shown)


main.go
```diff
@@ -66,6 +67,8 @@ func main() {
 
 	version := flag.Bool("version", false, "Display the version and exit")
 
+	debug := flag.Bool("enable-debug-endpoint", false, "Enable debug endpoint")
```
josselin-c, line 70: Remove the `enable` prefix: `--debug-endpoint`

pkg/cache/cache.go
```diff
@@ -40,6 +41,8 @@ type CacheResponse struct {
 type ServiceAccountCache interface {
 	Start()
 	Get(name, namespace string) (role, aud string, useRegionalSTS bool)
+	// List returns cache contents as JSON string
+	List() string
```
josselin-c, line 45: I would call this `ToJSON` or `MarshalJSON` or something if we systematically expect json. Otherwise `String() string` could also work. With `List()` I am expecting a slice, which is not what is returned.

pkg/cache/cache.go
```diff
@@ -79,6 +82,18 @@ func (c *serviceAccountCache) pop(name, namespace string) {
 	delete(c.cache, namespace+"/"+name)
 }
 
+// Dump cache contents for debugginqg
+func (c *serviceAccountCache) List() string {
+	c.mu.Lock()
```
josselin-c, line 87: nit: could even be `RLock()/RUnlock()` as this is a RWMutex.

main.go
shravan-achar: Should we call this --enable-debugging-handlers instead similar to the flag supported by kubelet?
Ref: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/#options

pkg/cache/cache.go
shravan-achar: Reusing the lock used for other cache operations such as get and set to ensure synchronization, instead of creating another one.

pkg/cache/cache.go
shravan-achar: Makes sense. ToJSON seems to fit the return type.

main.go
josselin-c: sounds good to me

shravan-achar: Updated

main.go
```diff
@@ -66,6 +67,8 @@ func main() {
 
 	version := flag.Bool("version", false, "Display the version and exit")
 
+	debug := flag.Bool("enable-debugging-handlers", false, "Enable debugging handlers. Currently /debug/cache is supported")
```
josselin-c, line 70: /debug/alpha/cache

shravan-achar: Ahh yes. Will fix that.

josselin-c (approved):
lgtm

pkg/handler/handler.go
```diff
@@ -360,6 +373,8 @@ func (m *Modifier) MutatePod(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResp
 
 	// determine whether to perform mutation
 	if podRole == "" {
+		klog.V(3).Infof("Pod was not mutated. %s",
+			logContext(pod.Name, pod.GenerateName, pod.Spec.ServiceAccountName, pod.Namespace))
```
jqmichael, line 377: I think it will be more valuable if we log the reason why it's not muted.

jqmichael (commented):
What about other cases where pod is not mutated, i.e. no service account is found or annotated?

main.go
```diff
@@ -66,6 +67,8 @@ func main() {
 
 	version := flag.Bool("version", false, "Display the version and exit")
 
+	debug := flag.Bool("enable-debugging-handlers", false, "Enable debugging handlers. Currently /debug/alpha/cache is supported")
+
```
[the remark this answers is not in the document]
shravan-achar: Since its a debug feature, it was thought best to leave the default as false. Any alternate thoughts? cc: @josselin-c

pkg/handler/handler.go
shravan-achar: I will add more logging in the upcoming PR

PR #92, opened 2020-11-19
Add additional log statements and update client-go

shravan-achar:
*Issue #, if available:*

*Description of changes:*

- Add logging statements where containers/pod already have required properties
- Update client-go to kubernetes-1.18. Update to go1.13


go.mod
```diff
@@ -1,32 +1,20 @@
...
-	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
-	gopkg.in/inf.v0 v0.9.1 // indirect
+	github.com/spf13/pflag v1.0.5
 	gopkg.in/square/go-jose.v2 v2.5.1
-	k8s.io/api v0.0.0-20190606204050-af9c91bd2759
-	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
-	k8s.io/client-go v11.0.1-0.20190606204521-b8faab9c5193+incompatible
-	k8s.io/klog v0.3.0
-	k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208 // indirect
-	k8s.io/kubernetes v1.14.3
```
jaypipes, line 29: nice.

pkg/cert/request.go
```diff
@@ -65,8 +65,8 @@ func NewServerCertificateManager(kubeClient clientset.Interface, namespace, secr
 			// authenticate itself to a TLS client.
 			certificates.UsageServerAuth,
 		},
-		CertificateStore:      certificateStore,
-		CertificateExpiration: certificateExpiration,
+		CertificateStore:    certificateStore,
+		CertificateRotation: certificateRotation,
```
jaypipes, line 69: This change to `pkg/cert/request.go` seems to be unrelated to updating log messages and client-go. Would you mind pulling this out into its own separate PR please?

jaypipes (changes requested):
Great stuff @shravan-achar :) Just one request inline to remove the unrelated Gauge -> Histogram metric change.

Thanks!
-jay

pkg/cert/request.go
mcristina422: I ran into the same issue when updating client-go. This is related to the client-go change, the metric was replaced internally and we are forced to make the change here

mcristina422: PR here https://github.com/kubernetes/kubernetes/pull/85874

shravan-achar: The gauge metric seems to have been removed in client-go 1.18. Thanks @mcristina422 for the PR link. I was looking for it but couldn't find it.

pkg/handler/handler.go
```diff
@@ -373,14 +381,19 @@ func (m *Modifier) MutatePod(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResp
 
 	// determine whether to perform mutation
 	if podRole == "" {
-		klog.V(3).Infof("Pod was not mutated. %s",
-			logContext(pod.Name, pod.GenerateName, pod.Spec.ServiceAccountName, pod.Namespace))
+		klog.V(3).Infof("Pod was not mutated. Reason: %s. %s",
+			"Service account did not have the right annotations or was not found",
```
jqmichael, line 385: was not found from the cache.

pkg/handler/handler.go
```diff
@@ -149,18 +147,18 @@ func logContext(podName, podGenerateName, serviceAccountName, namespace string)
...
 		namespace)
 }
 
-func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, roleName string, podSettings *podUpdateSettings) {
+func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, roleName string, podSettings *podUpdateSettings) bool {
 	// return if this is a named skipped container
 	if _, ok := podSettings.skipContainers[container.Name]; ok {
-		return
+		return false
 	}
```
jqmichael, line 162: Should we log in these cases?

pkg/cache/cache.go
```diff
@@ -60,6 +60,7 @@ func (c *serviceAccountCache) Get(name, namespace string) (role, aud string, use
 	klog.V(5).Infof("Fetching sa %s/%s from cache", namespace, name)
 	resp := c.get(name, namespace)
 	if resp == nil {
+		klog.V(3).Infof("Service account %s/%s not found in cache", namespace, name)
```
jqmichael, line 63: Nice!

pkg/handler/handler.go
```diff
@@ -390,8 +403,16 @@ func (m *Modifier) MutatePod(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResp
...
 
-	klog.V(3).Infof("Pod was mutated. %s",
-		logContext(pod.Name, pod.GenerateName, pod.Spec.ServiceAccountName, pod.Namespace))
+	// TODO: klog structured logging can make this better
+	if changed {
+		klog.V(3).Infof("Pod was mutated. %s",
+			logContext(pod.Name, pod.GenerateName, pod.Spec.ServiceAccountName, pod.Namespace))
+	} else {
+		klog.V(3).Infof("Pod was not mutated. Reason: %s. %s",
+			"Required volume mounts and env variables were already present",
```
jqmichael, line 412: By reading the code, it looks like the reason is most likely the env variables are already present (missing volume alone won't mutate the pod).

pkg/cert/request.go
mcristina422: In my WIP branch here https://github.com/aws/amazon-eks-pod-identity-webhook/compare/master...mcristina422:updateK8sio?expand=1 I also had to change the managers clientFn

jqmichael (commented):
/lgtm. 

Minor comments.

Can we test this change against a 1.15 cluster to ensure backwards-compatibility?

pkg/cert/request.go
shravan-achar: I believe that would be necessary when we move to client-go for k8s 1.19. 

For the histogram metric above, I need to change the metric name and the help text, since the metric tracks certificate lifetime instead of expiration seconds. I will update that. 

I wonder if we should also update the README on this change on the metric. It seems to impact those who may have set alarms on the "server_expiration_seconds" gauge metric. Please let me know if there are concerns on this @jaypipes @mcristina422 @josselin-c @jqmichael @micahhausler

pkg/handler/handler.go
shravan-achar: Ahh yes. Will fix

pkg/handler/handler.go
shravan-achar: Ya. We can add a log here as well. Will fix

pkg/handler/handler.go
shravan-achar: I did not fully understand this. If volume is missing on the pod we do mutate the pod here - https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L308. There are similar lines for adding volume mounts on eligible containers here https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L242

pkg/handler/handler.go
shravan-achar: Updated

pkg/handler/handler.go
shravan-achar: Updated

pkg/cert/request.go
shravan-achar: Updated

shravan-achar:
> Can we test this change against a 1.15 cluster to ensure backwards-compatibility?

@jqmichael - Can confirm the container image works against k8s 1.15 server

pkg/cert/request.go
jaypipes: Gotcha, thanks for the explanation @mcristina422 :)

jaypipes (approved):
/lgtm
/approve

pkg/handler/handler.go
nckturner: Nit: the reason is hard-coded, and yet using a string substitution "%s"...  It should just be one substitution for the logContext?

pkg/handler/handler.go
```diff
@@ -191,9 +190,12 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	}
 
 	if reservedKeysDefined && regionKeyDefined && regionalStsKeyDefined {
-		return
+		klog.V(3).Infof("container %s has necessary env variables already present",
```
nckturner, line 193: I think we might want to set the log level to be different for successes and skips.  I can see in some cases, someone might want to only log successes because the amount of logs generated by skips would be much larger, like 3 for success, 4 or 5 for skip?

pkg/cache/cache.go
```diff
@@ -60,6 +60,7 @@ func (c *serviceAccountCache) Get(name, namespace string) (role, aud string, use
 	klog.V(5).Infof("Fetching sa %s/%s from cache", namespace, name)
```
nckturner, line 60: Do we want a very high (i.e. 9) level log that prints out the service account cache on every read?

nckturner (commented):
Nice! Had some minor suggestions.

pkg/cache/cache.go
shravan-achar: I have been following this logging convention doc https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md. Anything >=5 is trace level logging. If I enable logLevel 9, I can see HTTP level logs (Request headers). So, I kept it at 5.

pkg/handler/handler.go
shravan-achar: Good point. I think we should separate them. As per the logging convention doc, I categorized them all as "Extended information about changes" which is V(3). There is only success log per pod, so it makes sense to add them at V(3) and rest as "Debug" info as V(4). Will fix.

pkg/handler/handler.go
shravan-achar: Ahh yes. Small change. Will fix.

shravan-achar:
@jaypipes @nckturner - Pushed a small update to bump up logging levels for some statements. Re-requesting review :)

pkg/handler/handler.go
jqmichael: My bad, was trying to point out missing "volume mounts" in the containers isn't enough to mutate since it will return early if env variables are set. It does get mutated if the "volumes" are missing.

https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L242

jqmichael (approved):

nckturner (approved):

Issue #56, opened 2020-05-06
Role ARN Prefix


dmolik:
**What would you like to be added**:

Add a CLI switch to allow for adding an IAM Role prefix. This feature should be overridden if a full ARN is detected in a ServiceAccount annotation. 

**Why is this needed**:

This allows for shorthand in service account annotations, further allowing, deployments to remain similar across accounts.

PR #95, opened 2020-11-21
Allow specifying baseArn to prepend to role names.

danp60:
*Description of changes:*
This PR introduces the ability to specify a baseArn to prepend to role names when we detect that the arn passed to the `eks.amazonaws.com/role-arn` annotation is not fully qualified. 

Our use case is that we have different clusters that run in different AWS accounts. This will allow us to use the same manifest to deploy to these different clusters and allow each pod to assume the correct IAM role local to their account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Usage of amazon-eks-pod-identity-webhook:
       --alsologtostderr                  log to standard error as well as files
       --annotation-prefix string         The Service Account annotation to look for (default "eks.amazonaws.com")
       --aws-default-region string        If set, AWS_DEFAULT_REGION and AWS_REGION will be set to this value in mutated containers
+      --base-arn string                  The base arn to use if a non fully qualified role is detected
       --in-cluster                       Use in-cluster authentication and certificate request API (default true)
       --enable-debugging-handlers        Enable debugging handlers on the metrics port (http). Currently /debug/alpha/cache is supported (default false) [ALPHA]
       --kube-api string                  (out-of-cluster) The url to the API server
--- a/main.go
+++ b/main.go
@@ -60,6 +60,7 @@ func main() {
 	// annotation/volume configurations
 	annotationPrefix := flag.String("annotation-prefix", "eks.amazonaws.com", "The Service Account annotation to look for")
 	audience := flag.String("token-audience", "sts.amazonaws.com", "The default audience for tokens. Can be overridden by annotation")
+	baseArn := flag.String("base-arn", "", "The base arn to use if a non fully qualified role is detected")
 	mountPath := flag.String("token-mount-path", "/var/run/secrets/eks.amazonaws.com/serviceaccount", "The path to mount tokens")
 	tokenExpiration := flag.Int64("token-expiration", 86400, "The token expiration")
 	region := flag.String("aws-default-region", "", "If set, AWS_DEFAULT_REGION and AWS_REGION will be set to this value in mutated containers")
@@ -112,6 +113,7 @@ func main() {
 		handler.WithServiceAccountCache(saCache),
 		handler.WithRegion(*region),
 		handler.WithRegionalSTS(*regionalSTS),
+		handler.WithBaseArn(*baseArn),
 	)
 
 	addr := fmt.Sprintf(":%d", *port)
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -74,6 +74,8 @@ var (
 	deserializer  = codecs.UniversalDeserializer()
 )
 
+const arnPrefix = "arn:"
+
 // ModifierOpt is an option type for setting up a Modifier
 type ModifierOpt func(*Modifier)
 
@@ -102,6 +104,11 @@ func WithRegionalSTS(enabled bool) ModifierOpt {
 	return func(m *Modifier) { m.RegionalSTSEndpoint = enabled }
 }
 
+// WithBaseArn sets the baseArn to use if we detect role name is not fully qualified
+func WithBaseArn(baseArn string) ModifierOpt {
+	return func(m *Modifier) { m.BaseArn = baseArn }
+}
+
 // WithAnnotationDomain adds an annotation domain
 func WithAnnotationDomain(domain string) ModifierOpt {
 	return func(m *Modifier) { m.AnnotationDomain = domain }
@@ -127,6 +134,7 @@ func NewModifier(opts ...ModifierOpt) *Modifier {
 // Modifier holds configuration values for pod modifications
 type Modifier struct {
 	AnnotationDomain    string
+	BaseArn             string
 	Expiration          int64
 	MountPath           string
 	Region              string
@@ -155,7 +163,7 @@ func logContext(podName, podGenerateName, serviceAccountName, namespace string) 
 		namespace)
 }
 
-func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, roleName string, podSettings *podUpdateSettings) bool {
+func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, fullRoleArn string, podSettings *podUpdateSettings) bool {
 	// return if this is a named skipped container
 	if _, ok := podSettings.skipContainers[container.Name]; ok {
 		klog.V(4).Infof("Container %s was annotated to be skipped", container.Name)
@@ -225,7 +233,7 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	if !reservedKeysDefined {
 		env = append(env, corev1.EnvVar{
 			Name:  "AWS_ROLE_ARN",
-			Value: roleName,
+			Value: fullRoleArn,
 		})
 
 		env = append(env, corev1.EnvVar{
@@ -258,6 +266,14 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	return changed
 }
 
+func (m *Modifier) fullRoleArn(arn string) string {
+	if strings.HasPrefix(arn, arnPrefix) {
+		return arn
+	}
+
+	return fmt.Sprintf("%s%s", m.BaseArn, arn)
+}
+
 func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string, regionalSTS bool) ([]patchOperation, bool) {
 	updateSettings := newPodUpdateSettings(m.AnnotationDomain, pod, regionalSTS)
 
@@ -272,17 +288,19 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string, reg
 		tokenFilePath = "C:" + strings.Replace(tokenFilePath, `/`, `\`, -1)
 	}
 
+	fullRoleArn := m.fullRoleArn(roleName)
+
 	var changed bool
 	var initContainers = []corev1.Container{}
 	for i := range pod.Spec.InitContainers {
 		container := pod.Spec.InitContainers[i]
-		changed = m.addEnvToContainer(&container, tokenFilePath, roleName, updateSettings)
+		changed = m.addEnvToContainer(&container, tokenFilePath, fullRoleArn, updateSettings)
 		initContainers = append(initContainers, container)
 	}
 	var containers = []corev1.Container{}
 	for i := range pod.Spec.Containers {
 		container := pod.Spec.Containers[i]
-		changed = m.addEnvToContainer(&container, tokenFilePath, roleName, updateSettings)
+		changed = m.addEnvToContainer(&container, tokenFilePath, fullRoleArn, updateSettings)
 		containers = append(containers, container)
 	}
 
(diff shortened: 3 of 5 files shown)


danp60:
Any feedback or comments @aws/eks-contributors?

danp60:
@jqmichael @josselin-c Any chance someone could take a look at this?

danopia:
I'd love to see this merged; it's silly to require the role annotation to differ between AWS accounts. Splitting accounts by environment is pretty common practice, as is one cluster per account.

Fixes #56

danp60:
@nckturner @wongma7 does this look like a worthwhile contribution to you? Please let me know and I will cleanup this PR

main.go
```diff
@@ -60,6 +60,7 @@ func main() {
 	// annotation/volume configurations
 	annotationPrefix := flag.String("annotation-prefix", "eks.amazonaws.com", "The Service Account annotation to look for")
 	audience := flag.String("token-audience", "sts.amazonaws.com", "The default audience for tokens. Can be overridden by annotation")
+	baseArn := flag.String("base-arn", "", "The base arn to use if a non fully qualified role is detected")
```
jyotimahapatra, line 63: I have few thoughts on this:
1- This mechanism allows usage only when the user has access to change the parameters. Non cluster admins would not be able to use this functionality. wdyt about making this extensible by adding an new annotations on service accounts?
2- rollout and rollback(if needed) of this might become tricky due to using the same rolearn annotation. wdyt about using entirely different parameters for annotation like `eks.amazonaws.com/role` and  `eks.amazonaws.com/role-prefix` that clarifies the intent of the fields explicitly
3. Can we add a feature flag for this functionality (set to alpha first and disabled by default) so that only interested users can flip this on.

johngmyers: The prefix is always going to be specific to the cluster. It will be `arn:aws:iam::`, followed by the ID of the account the cluster is in, followed by `:role/`, followed by the name of the cluster, followed by some fixed separator (I use `_`).

So one possibility would be to make this not configurable, simply prepending unqualified values with the prefix described above.

danp60: Thanks for the comments. I did not forget about this PR, and I'll try and make some time to work on it soon.

danp60: > 1- This mechanism allows usage only when the user has access to change the parameters. Non cluster admins would not be able to use this functionality. wdyt about making this extensible by adding an new annotations on service accounts?

One of the main purposes of this feature is to hide the details of the environment (ie the AWS account ID) from the end user. If the end user had to add an annotation with the base ARN that included the account ID, then they could just write the fully qualified ARN and not use this feature at all, so I'm not sure if it would be that useful.

> 2- rollout and rollback(if needed) of this might become tricky due to using the same rolearn annotation. wdyt about using entirely different parameters for annotation like eks.amazonaws.com/role and eks.amazonaws.com/role-prefix that clarifies the intent of the fields explicitly

I can do this. If an end user sets both, the most reasonable (and backwards compatible) behavior to me seems to be to take the original annotation `eks.amazonaws.com/role` verbatim and not make any changes.

> 3. Can we add a feature flag for this functionality (set to alpha first and disabled by default) so that only interested users can flip this on.

Do you think this is required if we do 2.  and make the new annotation only a fallback if the original one doesn't exist?

> The prefix is always going to be specific to the cluster. It will be arn:aws:iam::, followed by the ID of the account the cluster is in, followed by :role/, followed by the name of the cluster, followed by some fixed separator (I use _).
>
> So one possibility would be to make this not configurable, simply prepending unqualified values with the prefix described above.
This seems reasonable to me too. I can add a flag for --aws-default-account-id similar to how we have flags for other AWS environment like region.

WDYT?

seh:
Is anyone still working on this patch?

chlunde:
@jyotimahapatra would this approach help anyone running on EKS? I would expect some template like `{{clusterAccount}}` might be supported in EKS, but configuring a CLI option would only work for self-hosted installations

jyotimahapatra:
> would this approach help anyone running on EKS? I would expect some template like {{clusterAccount}} might be supported in EKS, but configuring a CLI option would only work for self-hosted installations

Correct. This is right now relevant for self hosted installations.

johngmyers:
How this would work on EKS is a problem for EKS. It is not particularly relevant for the amazon-eks-pod-identity-webhook project.

jyotimahapatra:
>How this would work on EKS is a problem for EKS. It is not particularly relevant for the amazon-eks-pod-identity-webhook project.

Correct. Once this is implemented, EKS can make it work by default.

README.md
```diff
@@ -133,6 +133,7 @@ Usage of amazon-eks-pod-identity-webhook:
       --alsologtostderr                  log to standard error as well as files
       --annotation-prefix string         The Service Account annotation to look for (default "eks.amazonaws.com")
       --aws-default-region string        If set, AWS_DEFAULT_REGION and AWS_REGION will be set to this value in mutated containers
+      --base-arn string                  The base arn to use if a non fully qualified role is detected
```
seh, line 136:
```suggestion
      --role-base-arn string                  The base ARN to use as a prefix for completing role ARNs that are not fully qualified
```

main.go
seh, line 63:
```suggestion
	roleBaseARN := flag.String("role-base-arn", "", "The base ARN to use as a prefix for completing role ARNs that are not fully qualified")
```

main.go
```diff
@@ -112,6 +113,7 @@ func main() {
 		handler.WithServiceAccountCache(saCache),
 		handler.WithRegion(*region),
 		handler.WithRegionalSTS(*regionalSTS),
+		handler.WithBaseArn(*baseArn),
```
seh, line 116:
```suggestion
		handler.WithRoleBaseARN(*roleBaseARN),
```

pkg/handler/handler.go
```diff
@@ -102,6 +104,11 @@ func WithRegionalSTS(enabled bool) ModifierOpt {
 	return func(m *Modifier) { m.RegionalSTSEndpoint = enabled }
 }
 
+// WithBaseArn sets the baseArn to use if we detect role name is not fully qualified
+func WithBaseArn(baseArn string) ModifierOpt {
```
seh, line 108:
```suggestion
// WithRoleBaseARN sets the base ARN to use as a prefix for completing role ARNs that are not fully qualified
func WithRoleBaseARN(baseARN string) ModifierOpt {
```

pkg/handler/handler.go
```diff
@@ -102,6 +104,11 @@ func WithRegionalSTS(enabled bool) ModifierOpt {
...
+	return func(m *Modifier) { m.BaseArn = baseArn }
```
seh, line 109:
```suggestion
	return func(m *Modifier) { m.RoleBaseARN = baseARN }
```

pkg/handler/handler.go
```diff
@@ -127,6 +134,7 @@ func NewModifier(opts ...ModifierOpt) *Modifier {
 // Modifier holds configuration values for pod modifications
 type Modifier struct {
 	AnnotationDomain    string
+	BaseArn             string
```
seh, line 137:
```suggestion
	RoleBaseARN             string
```

pkg/handler/handler.go
```diff
@@ -155,7 +163,7 @@ func logContext(podName, podGenerateName, serviceAccountName, namespace string)
 		namespace)
 }
 
-func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, roleName string, podSettings *podUpdateSettings) bool {
+func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, fullRoleArn string, podSettings *podUpdateSettings) bool {
```
seh, line 166:
```suggestion
func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath, fullRoleARN string, podSettings *podUpdateSettings) bool {
```

pkg/handler/handler.go
```diff
@@ -225,7 +233,7 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	if !reservedKeysDefined {
 		env = append(env, corev1.EnvVar{
 			Name:  "AWS_ROLE_ARN",
-			Value: roleName,
+			Value: fullRoleArn,
```
seh, line 236:
```suggestion
			Value: fullRoleARN,
```

pkg/handler/handler.go
```diff
@@ -258,6 +266,14 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	return changed
 }
 
+func (m *Modifier) fullRoleArn(arn string) string {
```
seh, line 269:
```suggestion
func (m *Modifier) fullRoleARN(arn string) string {
```

pkg/handler/handler.go
```diff
@@ -258,6 +266,14 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
...
+	if strings.HasPrefix(arn, arnPrefix) {
+		return arn
+	}
+
+	return fmt.Sprintf("%s%s", m.BaseArn, arn)
```
seh, line 274:
```suggestion
	return fmt.Sprintf("%s%s", m.RoleBaseARN, arn)
```

pkg/handler/handler.go
```diff
@@ -272,17 +288,19 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string, reg
 		tokenFilePath = "C:" + strings.Replace(tokenFilePath, `/`, `\`, -1)
 	}
 
+	fullRoleArn := m.fullRoleArn(roleName)
```
seh, line 291:
```suggestion
	fullRoleARN := m.fullRoleARN(roleName)
```

pkg/handler/handler.go
```diff
@@ -272,17 +288,19 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string, reg
...
+
 	var changed bool
 	var initContainers = []corev1.Container{}
 	for i := range pod.Spec.InitContainers {
 		container := pod.Spec.InitContainers[i]
-		changed = m.addEnvToContainer(&container, tokenFilePath, roleName, updateSettings)
+		changed = m.addEnvToContainer(&container, tokenFilePath, fullRoleArn, updateSettings)
```
seh, line 297:
```suggestion
		changed = m.addEnvToContainer(&container, tokenFilePath, fullRoleARN, updateSettings)
```

pkg/handler/handler.go
```diff
@@ -272,17 +288,19 @@ func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string, reg
...
 		initContainers = append(initContainers, container)
 	}
 	var containers = []corev1.Container{}
 	for i := range pod.Spec.Containers {
 		container := pod.Spec.Containers[i]
-		changed = m.addEnvToContainer(&container, tokenFilePath, roleName, updateSettings)
+		changed = m.addEnvToContainer(&container, tokenFilePath, fullRoleArn, updateSettings)
```
seh, line 303:
```suggestion
		changed = m.addEnvToContainer(&container, tokenFilePath, fullRoleARN, updateSettings)
```

pkg/handler/handler_pod_test.go
```diff
@@ -48,6 +48,7 @@ var (
 	handlerExpirationAnnotation = "testing.eks.amazonaws.com/handler/expiration"
 	handlerRegionAnnotation     = "testing.eks.amazonaws.com/handler/region"
 	handlerSTSAnnotation        = "testing.eks.amazonaws.com/handler/injectSTS"
+	handlerBaseArnAnnotation    = "testing.eks.amazonaws.com/handler/baseArn"
```
seh, line 51:
```suggestion
	handlerBaseARNAnnotation    = "testing.eks.amazonaws.com/handler/baseARN"
```

pkg/handler/handler_pod_test.go
```diff
@@ -73,6 +74,9 @@ func getModifierFromPod(pod corev1.Pod) (*Modifier, error) {
 		}
 		modifiers = append(modifiers, WithRegionalSTS(value))
 	}
+	if baseArnAnnotation, ok := pod.Annotations[handlerBaseArnAnnotation]; ok {
+		modifiers = append(modifiers, WithBaseArn(baseArnAnnotation))
```
seh, line 78:
```suggestion
	if baseARNAnnotation, ok := pod.Annotations[handlerBaseARNAnnotation]; ok {
		modifiers = append(modifiers, WithBaseARN(baseARNAnnotation))
```

pkg/handler/testdata/rawPodNoBaseArn.pod.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: balajilovesoreos
+  annotations:
+    testing.eks.amazonaws.com/handler/baseArn: 'arn:aws:iam::111122223333:role/'
```
seh, line 6:
```suggestion
    testing.eks.amazonaws.com/handler/baseARN: 'arn:aws:iam::111122223333:role/'
```

PR #97, opened 2020-12-03
Add annotation option for token expiration

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Added annotation capability for both ServiceAccount and Pod to enable option to modify the expirationSeconds for the token, just like it is available as a flag to the binary. 

Note that the annotation in the pod would take precedence over the annotation in the service account to allow the ability to have certain pods where a customer wants to use a different expiration in seconds. Mentioned this distinction in the README as well.

### DIFF
--- a/pkg/cache/fake.go
+++ b/pkg/cache/fake.go
@@ -5,6 +5,8 @@ import (
 	"k8s.io/api/core/v1"
 	"strconv"
 	"sync"
+
+	"github.com/aws/amazon-eks-pod-identity-webhook/pkg"
 )
 
 // FakeServiceAccountCache is a goroutine safe cache for testing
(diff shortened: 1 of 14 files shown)


pkg/handler/testdata/rawPodHasAll.pod.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     testing.eks.amazonaws.com/serviceAccount/audience: "sts.amazonaws.com"
     testing.eks.amazonaws.com/handler/injectSTS: "true"
     testing.eks.amazonaws.com/handler/region: "cn-north-1"
+    eks.amazonaws.com/token-expiration: "86400"
```
micahhausler, line 11: Can you add a second test that exercises the service account annotation? Also, can you use the non-default token expiration in a test?

pkg/cache/fake.go
```diff
@@ -37,24 +42,25 @@ var _ ServiceAccountCache = &FakeServiceAccountCache{}
...
 
 // Get gets a service account from the cache
-func (f *FakeServiceAccountCache) Get(name, namespace string) (role, aud string, useRegionalSTS bool) {
+func (f *FakeServiceAccountCache) Get(name, namespace string) (role, aud string, useRegionalSTS bool, tokenExpiration int64) {
 	f.mu.RLock()
 	defer f.mu.RUnlock()
 	resp, ok := f.cache[namespace+"/"+name]
 	if !ok {
-		return "", "", false
+		return "", "", false, 86400
```
micahhausler, line 50: Can you make 86400 a `const`?

micahhausler (changes requested):

pkg/cache/fake.go
vivek-koppuru: Yup was thinking that, will do

pkg/handler/testdata/rawPodHasAll.pod.yaml
vivek-koppuru: Done

micahhausler (approved):
LGTM

rittneje:
For anyone wishing to leverage this feature, please note that the minimum expiration time is one hour (i.e., "3600"). If you try to set an expiration time less than this, it will silently be reset to the minimum.

Issue #94, opened 2020-11-19
Support cert-manager based deployments 


mcristina422:
**What would you like to be added**: 

Enable cert-manager enabled clusters to use this webhook 

This request is to support webhook certificates via a https://github.com/jetstack/cert-manager ca injector. This will benefit the webhook in a few ways, the certificate will no longer need to be generated by the application itself, removing a dependency on the CSR based upstream k8s certificate manager (where issues do arise like the metrics changes in https://github.com/aws/amazon-eks-pod-identity-webhook/pull/92). Additionally, it will remove the need for secret and CSR RBAC, as cert-manager will [inject the CA](https://cert-manager.io/docs/concepts/ca-injector/) to the webhook at deploy time. The certificate will get rotated in place when it expires

**Why is this needed**:
The "standard" set by kubebuilder/controller-runtime is cert manager based certificates as it simplifies certificate management 
https://book.kubebuilder.io/cronjob-tutorial/cert-manager.html#deploying-the-cert-manager

Both cert-manager based webhooks and CSR webhooks can both be supported at the same time with feature flags and updated RBAC. We run this today internally by updating RBAC, creating a cert-manager.io Certificate, setting the cert-manager.io/inject-ca-from annotation on the webhook, disabling the in cluster flag, and manually setting tls-cert and tls-key to be a mounted secret volume. But I think the server needs to reload the certificate, like in https://github.com/kubernetes-sigs/controller-runtime/pull/421/files or else when the cert expires, the server will still use the certificate it loaded at startup, not the live certificate on the filesystem 

EDIT: The certwatcher may get exported in https://github.com/kubernetes-sigs/controller-runtime/issues/1292

avdhoot:
@mcristina422 Can you please share steps how did it?

avdhoot:
Thanks @mcristina422 for hints. Below yaml for how you can deploy `amazon-eks-pod-identity-webhook` with cert-manager
```apiVersion: apps/v1
kind: Deployment
metadata:
  name: pod-identity-webhook
  namespace: kube-system
spec:
  replicas: 1
  selector:
    matchLabels:
      app: pod-identity-webhook
  template:
    metadata:
      labels:
        app: pod-identity-webhook
    spec:
      serviceAccountName: pod-identity-webhook
      containers:
        - name: pod-identity-webhook
          image: amazon/amazon-eks-pod-identity-webhook:2c44edc
          imagePullPolicy: Always
          command:
[...]

PR #115, opened 2021-05-07
certificatesigningrequests/v1beta1 to v1

joelddiaz:
*Issue #, if available:*

*Description of changes:*

Kube 1.22 drops support for certificatesigningrequests/v1beta1. Update vendor/k8s.io/client-go and migrate to certificatesigningrequests/v1.


2uasimojo:
@jqmichael would you be able to (find some one to) review this? Thanks!

danquack:
is there any updates on this?

2uasimojo:
@micahhausler @nckturner ?

danquack:
@jyotimahapatra @wongma7 ?

danopia:
Kubernetes v1.23 is out now, so this project is now 2 minors behind the latest Kubernetes release.

The fact that the `system:node:` prefix is required means that something is being done incorrectly I believe.. the CSR API is intended for grabbing certificates that the whole cluster trusts, but webhooks don't need that level of certificate. They just need to have their `caBundle` put into the `MutatingWebhookConfiguration` resource. It's still good IMO to get csr/v1 merged to work on newer Kubernetes but there will be future work to be a better Kubernetes citizen and not need `kubectl certificate approve` just to register a webhook.

h3poteto:
> but webhooks don't need that level of certificate. They just need to have their caBundle put into the MutatingWebhookConfiguration resource. 

I agree. I think it's good to support cert-manager based deployments in the future. Refs: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/94

h3poteto:
> With this patch the CSRs for the pod-identity-webhook are now created (where as before they failed to be created at all).
However when you go to approve them they end up in an Approved,Failed state:

I have the same issue, and it is solved by [that change](https://github.com/aws/amazon-eks-pod-identity-webhook/pull/115#issuecomment-951460258). The `system:node` prefix is required as long as we continue to use the `kubernetes.io/kubelet-serving ` as a signer.
So could you please add that change in this PR, @joelddiaz ? Or shall I do it with a new PR?

wongma7:
Yes it seems like requesting kubelet-serving with system:node prefix is the path of least resistance for maintaining roughly the same functionality/behavior in 1.22 as if `legacy-unknown` were allowed. So I am inclined to merge this with the system:node prefix change. I understand the sentiment about it being "wrong", it does feel a bit to me like abuse of the API since obviously the webhook is not a kubelet. But it seems also like other solutions require third party software and/or more significant changes.

It might be "wrong" for the webhook to be getting its own serving cert in the first place...it is mimicking what kubelet does. From what I have seen, it is more common a pattern to have the cert generation be the responsibility of the deployer as a prerequisite to webhook deployment. So for example, another aws maintained webhook https://github.com/kubernetes-sigs/aws-iam-authenticator has an optional step where certs are generated *before* deployment. As mentioned in https://github.com/aws/amazon-eks-pod-identity-webhook/issues/94 if the webhook could just read the cert from file and reload then it could be ignorant of how certs are generated, by cert-manager, manually or whatever.

Another complication is that AWS EKS does NOT run the kubernetes default signer. So having the webhook create a kublet-serving cert might not work on EKS. But EKS 1.22 is not released yet so that's not really a problem yet

h3poteto:
> But EKS 1.22 is not released yet so that's not really a problem yet

Oh, sure...

wongma7:
:joy: don't worry, it's on AWS's radar to fix the problem *before* release. 

If you have time to submit a PR with the change if @joelddiaz is not available (to be fair, we took very long to review this PR!) I can approve it. Thanks

wongma7:
Actually I have time to submit a PR now, will link to it in a minute and appreciate reviews. thx.

wongma7:
https://github.com/aws/amazon-eks-pod-identity-webhook/pull/133

wongma7:
#133 is up for review , (sorry for noise, I sent previous comment with just hte link )

and btw when I said 'not a problem' yet I meant for EKS users specifically. Obviously this repository has a wider audience and I'll try to prioritize the cert-manager fix even if EKS doesn't necessarily officially support that as an add-on or whatever.

wongma7:
Solution I'm thinking of is (cross-posting from https://github.com/aws/amazon-eks-pod-identity-webhook/pull/133#issuecomment-1017907120 for awareness, apologies again for all the noise):
- if --in-cluster=true then you must provide signerName (and common name and org, as new arguments). So for non-EKS users they can set these to kubelet-serving and system:node accordingly. And for other distros that have their custom signer they might support other signerNames and such.
- if --in-cluster=false then #134 should fix the bug where cert never gets reloaded if it's renewed so that fixes the cert-manager use-case

danquack:
Looks like AWS released EKS 1.22 3 days ago, but noticed this did not make it. Is a different work around for getting this deployed with 1.22? Not sure if you would know @wongma7

clayrosenthal:
Any update on this? The solution below seems pretty reasonable to me

> Solution I'm thinking of is (cross-posting from [#133 (comment)](https://github.com/aws/amazon-eks-pod-identity-webhook/pull/133#issuecomment-1017907120) for awareness, apologies again for all the noise):
> 
>     * if --in-cluster=true then you must provide signerName (and common name and org, as new arguments). So for non-EKS users they can set these to kubelet-serving and system:node accordingly. And for other distros that have their custom signer they might support other signerNames and such.
> 
>     * if --in-cluster=false then [Use certwatcher to support mounting cert-manager certificates. #134](https://github.com/aws/amazon-eks-pod-identity-webhook/pull/134) should fix the bug where cert never gets reloaded if it's renewed so that fixes the cert-manager use-case

danopia:
> Any update on this? The solution below seems pretty reasonable to me

I believe that everything's in place for this. The deployment YAML in https://github.com/prasita123/amazon-eks-pod-identity-webhook/tree/master/deploy has been updated to use cert-manager with `--in-cluster=false` and I've been running this configuration in production for two years.

PR #116, opened 2021-05-20
Correctly set the sts regional endpoint

jyotimahapatra:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -36,7 +36,7 @@ func TestSaCache(t *testing.T) {
 
 	assert.Equal(t, roleArn, role, "Expected role to be %s, got %s", roleArn, role)
 	assert.Equal(t, "sts.amazonaws.com", aud, "Expected aud to be sts.amzonaws.com, got %s", aud)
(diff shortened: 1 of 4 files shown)


pkg/cache/cache_test.go
```diff
@@ -32,16 +33,33 @@ func TestSaCache(t *testing.T) {
...
+	roleArn := "arn:aws:iam::111122223333:role/s3-reader"
+	testSA.Annotations = map[string]string{
+		"eks.amazonaws.com/role-arn":               roleArn,
+		"eks.amazonaws.com/sts-regional-endpoints": "false",
+		"eks.amazonaws.com/token-expiration":       "3600",
 	}
-	if tokenExpiration != 3600 {
-		t.Errorf("Expected token expiration to be 3600, got %d", tokenExpiration)
+
+	cache := &serviceAccountCache{
```
wongma7, line 54: IMO we need to test every case. 

1. annotation missing
2. annotation true
3. annotatoin false
4. annotation invalid
5. 1-4 when defaultRegionalSTS is set true
6. 1-4 when defaultRegionalSTS is set false

you can pass a fake client to New and put the testSa in it.

PR #120, opened 2021-08-11
Fix serviceaccount regional sts annotation not taking effect unless flag is true

wongma7:
*Issue #, if available:*

*Description of changes:*
first commit is the fix. spotted by  jyotimahapatra . Second commit is some refactoring that follows from the fix. basically it is unclear what parameters are coming from where, so the changes are to make the code clearer without changing behavior:

- Remove Expiration and RegionalSTSEndpoint from Modifier because they are
unused and redundant since serviceaccount cache already parses them.

- Refactor newPodUpdateSettings to parsePodAnnotations

- Rename updatePodSpec to getPodSpecPatch

- Fix changed being set false if the last container is skipped and no
other changes are made

I will test the changes on my EKS cluster

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -199,7 +199,7 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	changed := false
 	env := container.Env
 
-	if !regionalStsKeyDefined && m.RegionalSTSEndpoint && podSettings.useRegionalSTS {
+	if !regionalStsKeyDefined && podSettings.useRegionalSTS {
 		env = append(env,
 			corev1.EnvVar{
 				Name:  stsKey,
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -36,34 +36,6 @@ import (
 	"k8s.io/klog"
 )
 
-type podUpdateSettings struct {
-	skipContainers map[string]bool
-	useRegionalSTS bool
-}
-
-// newPodUpdateSettings returns the update settings for a particular pod
-func newPodUpdateSettings(annotationDomain string, pod *corev1.Pod, useRegionalSTS bool) *podUpdateSettings {
-	settings := &podUpdateSettings{
-		useRegionalSTS: useRegionalSTS,
-	}
-
-	skippedNames := map[string]bool{}
-	skipContainersKey := annotationDomain + "/" + pkg.SkipContainersAnnotation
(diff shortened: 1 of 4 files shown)


wongma7:
test results match unit tests https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/cache/cache_test.go#L63-L108

--sts-regional-endpoints=false, annotation true: AWS_STS_REGIONAL_ENDPOINTS=regional
--sts-regional-endpoints=false, annotation false: AWS_STS_REGIONAL_ENDPOINTS is omitted
--sts-regional-endpoints=false, no annotation: AWS_STS_REGIONAL_ENDPOINTS is omitted

--sts-regional-endpoints=true, annotation true: AWS_STS_REGIONAL_ENDPOINTS=regional
--sts-regional-endpoints=true, annotation false: AWS_STS_REGIONAL_ENDPOINTS is omitted
--sts-regional-endpoints=true, no annotation: AWS_STS_REGIONAL_ENDPOINTS=regional

jyotimahapatra (approved):
lgtm

pkg/handler/handler.go
```diff
@@ -259,9 +230,31 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 	return changed
 }
 
-func (m *Modifier) updatePodSpec(pod *corev1.Pod, roleName, audience string, regionalSTS bool, tokenExpiration int64) ([]patchOperation, bool) {
-	updateSettings := newPodUpdateSettings(m.AnnotationDomain, pod, regionalSTS)
+// parsePodAnnotations parses the pod annotations that can influence mutation:
+// - tokenExpiration. Overrides the given service account annotation/flag-level
```
nckturner, line 234: Nit: this ship probably already sailed, but should the service account annotation for token expiration really be overridable by the pod?  I.e. should someone with only permissions to the pod, but not the service account, be allowed to increase the expiration (vs. only being allowed to lower it)?

nckturner (approved):

pkg/handler/handler.go
wongma7: they already can by manually editing the pod. but the whole point of the webhook is so you dont have to manually edit anything.

I get what youre saying though in the sense that I dont really understand the logic why this in particular is controlled at pod level whereas the rest are serviceaccount/flag. at least with the comments it should be easier to navigate. because as it stands there are basically 4 * flags + 4 * sa annotations + 1 * pod annotation

PR #122, opened 2021-08-16
add metrics for knowing adoption

jyotimahapatra:
*Issue #, if available:*

*Description of changes:*
While handling the infrastructure for a cluster, we need a way to know which clusters are using the webhook to drive changes that change default values through the infra.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -64,6 +64,10 @@ var webhookUsage = prometheus.NewGauge(prometheus.GaugeOpts{
 	Help: "Indicator to know pod identity webhook is used",
 })
 
+func init() {
+	prometheus.MustRegister(webhookUsage)
+}
+
 func (c *serviceAccountCache) Get(name, namespace string) (role, aud string, useRegionalSTS bool, tokenExpiration int64) {
 	klog.V(5).Infof("Fetching sa %s/%s from cache", namespace, name)
 	resp := c.get(name, namespace)
@@ -151,6 +155,7 @@ func New(defaultAudience, prefix string, defaultRegionalSTS bool, defaultTokenEx
 		defaultRegionalSTS:     defaultRegionalSTS,
 		defaultTokenExpiration: defaultTokenExpiration,
 		hasSynced:              informer.Informer().HasSynced,
+		webhookUsage:           webhookUsage,
 	}
 
 	informer.Informer().AddEventHandler(
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg"
+	"github.com/prometheus/client_golang/prometheus"
 	v1 "k8s.io/api/core/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	coreinformers "k8s.io/client-go/informers/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/cache/cache.go
```diff
@@ -163,6 +174,10 @@ func New(defaultAudience, prefix string, defaultRegionalSTS bool, defaultTokenEx
 						return
 					}
 				}
+				if _, ok := sa.Annotations[c.annotationPrefix+"/"+pkg.RoleARNAnnotation]; ok {
```
wongma7, line 177: the indenting looks very weird here, is it tabs vs spaces? gofmt?

wongma7: also is it intentional to call Dec twice?

pkg/cache/cache.go
```diff
@@ -122,6 +124,7 @@ func (c *serviceAccountCache) addSA(sa *v1.ServiceAccount) {
 				resp.TokenExpiration = pkg.ValidateMinTokenExpiration(tokenExpiration)
 			}
 		}
+		c.saCount.Inc()
```
wongma7, line 127: addSA will get called every time a service account is updated, created, or even anytime the informer cache gets resynced (every 60s https://github.com/aws/amazon-eks-pod-identity-webhook/blob/019d1ba3137fc48ce5c75f1e5b0f1f5dd679aed7/main.go#L101) so this number will quickly become inaccurate. Any way to emit the metric at a lower layer (in set? )

wongma7: maybe it's easier to emit a metric for when the webhook performs a mutation?

jyotimahapatra: makes sense..made the change

pkg/cache/cache.go
jyotimahapatra: changed the approach.

pkg/handler/handler.go
```diff
@@ -41,6 +42,15 @@ type podUpdateSettings struct {
 	useRegionalSTS bool
 }
 
+var mutations = prometheus.NewCounter(prometheus.CounterOpts{
+	Name: "mutations",
```
wongma7, line 46: by convention the name should have "_total" suffix:
...should have a suffix describing the unit, in plural form. Note that an accumulating count has total as a suffix, in addition to the unit if applicable. 
https://prometheus.io/docs/practices/naming/

wongma7 (approved):
one comment about metric name lgtm otherwise

wongma7 (dismissed):

wongma7 (approved):

nckturner (dismissed):
Would an equivalent metric that contains more information be a better alternative, like number of pods injected, number of service accounts with annotations seen, etc?  If `num_annotated_service_accounts_seen = 0` then we can assume the cluster is not using the webhook, but I'm not sure a good way to actually keep track of that.  A set of service account UIDs would work but would grow unbounded.  A gauge that returned the current number of annotated service accounts would work but wouldn't show clusters that had seen previous usage (not sure if we want that).  Maybe you can add some context to the description, and if this is just the simplest option for the information we need, I'm ok with that.

pkg/cache/cache.go
```diff
@@ -53,6 +54,18 @@ type serviceAccountCache struct {
 	defaultAudience        string
 	defaultRegionalSTS     bool
 	defaultTokenExpiration int64
+	webhookUsage           prometheus.Gauge
+}
+
+// We need a way to know if the webhook is used in a cluster to drive changes.
+// We could perform more interesting operations by knowing how many service accounts are being annotated.
```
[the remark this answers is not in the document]
jyotimahapatra: I brainstormed a bit about this. The number of pods injected is a good metrics, but that doesn't give us details about usage if the pods don't churn. We could add that as well, but we decided we don't have a usecase for the information.
The number of annotated service accounts is a good metric, but we it needs a bit of refactoring to make sure resync and update don't cause noise in the number. We chose to stick closer to what we need, know about usage in the clusters we operate.

nckturner (approved):

PR #134, opened 2022-01-20
Use certwatcher to support mounting cert-manager certificates.

colinhoglund:
*Issue #:*

Closes https://github.com/aws/amazon-eks-pod-identity-webhook/issues/94

*Description of changes:*

Mounting TLS certs from k8s Secret managed by https://cert-manager.io/ is a common approach for k8s webhooks.

Changes:
- Updates the "out-of-cluster" configuration to to use [certwatcher](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/certwatcher) for watching and updating the configured TLS cert/key.
- Moves signal handling out of the `ShutdownOnTerm` function and into `main` so that the context can be sent to both `certwatcher` and the `http.Server`.
- Updates k8s dependencies in order to use the certwatcher package, which moves some of the types in `pkg/cert/request.go` from `certificates/v1beta1` to `certificates/v1`.


pkg/cert/request.go
```diff
@@ -65,7 +64,7 @@ func NewServerCertificateManager(kubeClient clientset.Interface, namespace, secr
 			// authenticate itself to a TLS client.
 			certificates.UsageServerAuth,
 		},
-		SignerName:          certificates.LegacyUnknownSignerName,
+		SignerName:          "kubernetes.io/legacy-unknown",
```
colinhoglund, line 67: Hard coding this string since `LegacyUnknownSignerName` is no longer available in `v1`, as per https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers.

pkg/cert/request.go
```diff
@@ -65,7 +64,9 @@ func NewServerCertificateManager(kubeClient clientset.Interface, namespace, secr
 			// authenticate itself to a TLS client.
 			certificates.UsageServerAuth,
 		},
-		SignerName:          certificates.LegacyUnknownSignerName,
+		// Hard coding this since LegacyUnknownSignerName is no longer available in certificates/v1
+		// https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers.
+		SignerName:          "kubernetes.io/legacy-unknown",
```
wongma7, line 69: it's not just unavailable, it's not allowed: https://github.com/kubernetes/kubernetes/blob/48da959dbff18bfef6e801bd8c8ab3c88b7a7650/pkg/apis/certificates/validation/validation.go#L202

I think we will need to deprecate in-cluster flag + print a clear error + exit gracefully if the kubernetes server version is 1.22 AND `in-cluster` is true. Instead of trying and failing to create invalid CSR. However, doing all that is out of scope of your PR. So I would like to accept the other changes and drop these changes to ` pkg/cert/request.go`, if possible.

wongma7:
Thank you very much for this PR. I have one comment about the v1 CSR change . I think we will need to handle that separately by deprecating the in-cluster=true flag altogether because even if you set legacy-unknown it will be rejected for v1 CSRs. But open to discussion about that.

pkg/cert/request.go
colinhoglund: Thanks for the feedback @wongma7, and it totally makes sense to separate these. The only reason I changed `pkg/cert/request.go` is because of dependencies. `certwatcher` requires a newer version of the `controller-runtime`, which resulted in a upgrade of `k8s.io/client-go` to `v0.23.0`, which is turn changed the underlying [Config type](https://pkg.go.dev/k8s.io/client-go/util/certificate#Config) for `KeyUsage` to `certficates/v1`.

I'm not entirely sure if there is a way to avoid these dependency updates for now, but would be open to any suggesstions.

wongma7: I see, thx for the explanation. Then I agree with merging this part as well, I don't see a way to avoid it either.

wongma7: Note for self and others interested in the `in-cluster=true` case:

Given that this PR basically solves the `in-cluster=false` case since now webhook actually reload certs when they are renewed...

My plan for `in-cluster=true` is...

1. Currently, `in-cluster=true` works for k8s <=1.21 but doesn't for >1.21. We have discussed this at length and there seems to be no way to make it work in >1.22 (without mis/abusing the CSR api) so we need to deprecate it ASAP (as 1.22 has been out for months now).
2. Upon merging this, `in-cluster=true` won't work for ANY k8s version. Users who rely on it will need to make sure they use an older revision when building pod-identity-webhook.
3. Immediately/soon after merging this, someone (me/colleague/contributor) should deprecate `in-cluster=true` with a note in README and major version release so it is clear & obvious that it won't work for ANY k8s version

wongma7 (approved):

wongma7:
I dont actually have  permission in this repo so pinging others for review @nckturner etc. thanks !!

pkg/cert/request.go
colinhoglund: > 2\. Upon merging this, `in-cluster=true` won't work for ANY k8s version. Users who rely on it will need to make sure they use an older revision when building pod-identity-webhook.

It probably makes sense to tag a release before merging this since the latest release is a bit old. Then once the followup work is done, a new release can be cut that introduces `certwatcher` and deprecates the `--in-cluster` flag.

nckturner:
@prasita123 was working on testing these changes with a cert-manager deployment, can you update the results of your test?

prasita123:
tested the changes from this PR along with updates to 
- **deploy/deployment-base.yaml** 
    - setting in-cluster=false
    - updates to volumeMounts paths
    - add a cert-manager deployment
    - a cert-manager.io/v1  ClusterIssuer 
    - a cert-manager.io/v1 Certificate
 - **deploy/mutatingwebhook.yaml**
    - add annotations to replace caBundle
    
 Works as expected, cert-rotation was observed.

nckturner:
Failing dependency when go test is run:
```
 Error: ../../../go/pkg/mod/sigs.k8s.io/json@v0.0.0-20211020170558-c049b76a60c6/internal/golang/encoding/json/encode.go:1249:12: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
Error: ../../../go/pkg/mod/sigs.k8s.io/json@v0.0.0-20211020170558-c049b76a60c6/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
```

nckturner:
@colinhoglund can you take a look at the dependency failure?  Is it the go version we are running the tests with?

colinhoglund:
Thanks @nckturner, it looks like a dependency is using a feature of reflect added in 1.17: https://pkg.go.dev/reflect#StructField.IsExported

I've updated to go 1.17 and run `go mod vendor` and `go mod tidy`. Tests seem to be passing for me locally, but probably need to trigger a new run.

nckturner (approved):

colinhoglund:
> One more:
> 
> ```
>  Error: vendor/k8s.io/client-go/discovery/discovery_client.go:32:2: cannot find package "." in:
> 	/home/runner/work/amazon-eks-pod-identity-webhook/amazon-eks-pod-identity-webhook/vendor/github.com/googleapis/gnostic/openapiv2
> Error: Process completed with exit code 1.
> ```

Looks like this is a [case-sensitivity issue](https://github.com/kubernetes/client-go/issues/741) that I did not pick up locally (using darwin). I'm not entirely sure how to address it, but I'll take a look.

colinhoglund:
> > One more:
> > ```
> >  Error: vendor/k8s.io/client-go/discovery/discovery_client.go:32:2: cannot find package "." in:
> > 	/home/runner/work/amazon-eks-pod-identity-webhook/amazon-eks-pod-identity-webhook/vendor/github.com/googleapis/gnostic/openapiv2
> > Error: Process completed with exit code 1.
> > ```
> 
> Looks like this is a [case-sensitivity issue](https://github.com/kubernetes/client-go/issues/741) that I did not pick up locally (using darwin). I'm not entirely sure how to address it, but I'll take a look.

It looks like I was able to get it to recognize the case change by committing `rm -rf vendor/github.com/googleapis/gnostic` and then rerunning and committing `go mod vendor`. 🤞

nckturner:
Awesome.  This looks good, I think we just need to update the README with the relevant information about this being somewhat breaking.  Also cut a release.. which I can get started on.  Do you want to add a summary to the README @colinhoglund?

wongma7:
imo this doesn't need readme update, this is fixing a bug. anyway the readme is very 'in-cluster=true' focused and i think fixing that is out of scope of this PR

nckturner:
I mean I think it needs one, since in-cluster=true is essentially deprecated now.  But we can merge that separately.

nckturner (approved):

PR #138, opened 2022-01-22
Add warning for --in-cluster=true to README

nckturner:
*Description of changes:*

Add warning about `--in-cluster=true` to README. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@
 
 This webhook is for mutating pods that will require AWS IAM access.
 
+## Note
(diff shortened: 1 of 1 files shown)


wongma7:
nice. 

TODO is: code change to say it's deprecated...and if it's deprecated it doesn't make much sense to be default true either.

even if we reimplement it with v1 CSR API support, requiring signerName etc., it should come back in a different name

wongma7 (approved):

jaypipes (approved):
:+1: good wording, thanks @nckturner

PR #142, opened 2022-02-09
Add option to pass configmap with mapping between SA and IAM role

olemarkus:
*Description of changes:*

This PR adds an option for the webhook to watch a configmap for additional SA to Role mappings. This is useful where there are tooling that creates IAM Roles and already know what SA should use them. In particular kOps already has this mapping. Adding the annotation to the SAs is then just additional manual work.

A few notes:
* There is a number of ways to watch a configmap. I opted for using another informer that is namespaced rather than e.g mounting it on disk and watching the file.
* Previously, all SAs were added to cache regardless of if they had the annotation. This made an empty SA override what was populated from the ConfigMap if the SA was created/changed after the CM was read. I changed this to only add to cache if the annotation is present. I believe this doesn't change much as the webhook need to handle an SA that is not in cache anyway.
* The struct used in the configmap is the same as `CacheResponse`. This takes an orignally internal struct and makes it into a public interface. Mostly something to be aware of if one wants to change this struct in the future.

/cc @nckturner 

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,11 +143,12 @@ When running a container with a non-root user, you need to give the container ac
 
 ```
 Usage of amazon-eks-pod-identity-webhook:
+      --add_dir_header                   If true, adds the file directory to the header
       --alsologtostderr                  log to standard error as well as files
       --annotation-prefix string         The Service Account annotation to look for (default "eks.amazonaws.com")
       --aws-default-region string        If set, AWS_DEFAULT_REGION and AWS_REGION will be set to this value in mutated containers
+      --enable-debugging-handlers        Enable debugging handlers. Currently /debug/alpha/cache is supported
       --in-cluster                       Use in-cluster authentication and certificate request API (default true)
-      --enable-debugging-handlers        Enable debugging handlers on the metrics port (http). Currently /debug/alpha/cache is supported (default false) [ALPHA]
       --kube-api string                  (out-of-cluster) The url to the API server
       --kubeconfig string                (out-of-cluster) Absolute path to the API server kubeconfig file
       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
@@ -156,11 +157,11 @@ Usage of amazon-eks-pod-identity-webhook:
       --log_file_max_size uint           Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
       --logtostderr                      log to standard error instead of files (default true)
       --metrics-port int                 Port to listen on for metrics and healthz (http) (default 9999)
-      --namespace string                 (in-cluster) The namespace name this webhook and the tls secret resides in (default "eks")
+      --namespace string                 (in-cluster) The namespace name this webhook, the TLS secret, and configmap resides in (default "eks")
       --port int                         Port to listen on (default 443)
       --service-name string              (in-cluster) The service name fronting this webhook (default "pod-identity-webhook")
       --skip_headers                     If true, avoid header prefixes in the log messages
-      --skip_log_headers                 If true, avoid headers when openning log files
+      --skip_log_headers                 If true, avoid headers when opening log files
       --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
       --sts-regional-endpoint false      Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to false.
       --tls-cert string                  (out-of-cluster) TLS certificate file path (default "/etc/webhook/certs/tls.crt")
@@ -172,6 +173,7 @@ Usage of amazon-eks-pod-identity-webhook:
   -v, --v Level                          number for the log level verbosity
       --version                          Display the version and exit
       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
+      --watch-config-map                 Enables watching serviceaccounts that are configured through the pod-identity-webhook configmap instead of using annotations
 ```
 
 ### AWS_DEFAULT_REGION Injection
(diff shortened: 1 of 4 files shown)


nckturner:
> Previously, all SAs were added to cache regardless of if they had the annotation. This made an empty SA override what was populated from the ConfigMap if the SA was created/changed after the CM was read. I changed this to only add to cache if the annotation is present. I believe this doesn't change much as the webhook need to handle an SA that is not in cache anyway.

What is the behavior if an annotation is removed and the configmap is not being used?

Also is it intended that someone can use both configmap and annotations (I assume so based on the PR).  In that case, which takes precedence?

nckturner:
> Easiest way to change this is probably that addSAalso takes in the old object and if the old object has the annotation, but the new one doesn't, we pop the SA from cache. This is how I solved an SA being removed from the configmap as well.

> Hopefully the two will match. Technically the last one takes precedence, but documentation-wise I would say that if the two differ, the behavior is undocumented. I have no idea what the user intent would be in such a situation either.

> I do want to support both configmap and annotation in parallel both to preserve existing configurations, and because annotations offer more flexibility for those who really need to fine-tune the behavior.

Understandable that you would want to support both, but I'm wondering—if we support hybrid environments, what happens if we have a case where an annotation or configmap entry is removed, but the alternative still exists?  Even if we look at the old and new object of a certain type, it seems like we can get into the following states: 

1. SA annotation removed, configmap still has entry, `CacheResponse` removed from cache, webhook no longer mutates pods for SA.
2. Configmap entry removed, SA still has annotation, `CacheResponse` removed from cache, webhook no longer mutates pods for SA.
3. SA not annotated, configmap entry added, `CacheResponse` added to cache, webhook mutates pods for SA.
4. Configmap does not have entry, SA annotation added, `CacheResponse` added to cache, webhook mutates pods for SA.  

It seems to me that the only difference between 1 and 3 is the order in which prior events took place, and the same is true for 2 and 4.  If my understanding is correct, and these would result in different webhook behavior, then I think we would need to solve this before allowing both modes to operate at the same time.  I think we would adopt a strict precedence (not call it undefined) like service account always wins, and we might need to cache both types separately so we can see the state of both types when making decisions (service account annotation exists, but configmap doesn't => use mapping from service account annotation).

olemarkus:
Thanks for the feedback. Had hoped to avoid having two caches, but seems like that is the best way to go about this. Added this now, plus a test that confirms the presedence order is as expected.

pkg/cache/cache.go
```diff
@@ -181,17 +223,70 @@ func New(defaultAudience, prefix string, defaultRegionalSTS bool, defaultTokenEx
...
+						utilruntime.HandleError(err)
+					}
+				},
+			},
+		)
+	}
 	return c
 }
 
+func (c *serviceAccountCache) populateCacheFromCM(oldCM, newCM *v1.ConfigMap) error {
```
wongma7, line 255: I don't think it is deterministic the end result of the cache if two Updates or an Update and an Add (at the 60s resync period) happen at the same time, it depends on the order of pushes/pops. Of course then the cache should get fixed again at the next 60s resync so it's not a big issue but a queue would avoid it.

main.go
```diff
@@ -68,6 +69,7 @@ func main() {
 	tokenExpiration := flag.Int64("token-expiration", pkg.DefaultTokenExpiration, "The token expiration")
 	region := flag.String("aws-default-region", "", "If set, AWS_DEFAULT_REGION and AWS_REGION will be set to this value in mutated containers")
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
+	watchConfigMap := flag.Bool("watch-config-map", false, "Whether to watch the pod-identity-webhook ConfigMap for additional Service Accounts. Defaults to `false`.")
```
wongma7, line 72: Could mention that it's in namespace `--namespace` and description for `--namespace` should also  be updated since it's being overloaded/reused now

wongma7 (approved):
2 suggestions, feel free to implement or ignore, i am fine with merging this as-is

main.go
nckturner: Maybe also add "these service accounts will be watched even if they do not have annotations"

pkg/cache/cache.go
```diff
@@ -73,36 +75,61 @@ func init() {
 
 func (c *serviceAccountCache) Get(name, namespace string) (role, aud string, useRegionalSTS bool, tokenExpiration int64) {
```
nckturner, line 76: Can we add a comment for `Get()` and also something to the README that explains the ConfigMap and precedence between CM/SA?

pkg/cache/cache.go
olemarkus: Doesn't the informer framework use an underlying channel? So they should run sequentially, if I understand this correctly, not concurrently (no go routines).

wongma7: no you're right, the informer events are ordered and blocking, not sure where i got the impression otherwise but educated myself :woman_facepalming: https://github.com/kubernetes/client-go/blob/v0.23.0/tools/cache/shared_informer.go#L34

main.go
olemarkus: I tried to add something here. Kept it brief so that the usage docs for the relevant flags do not get too long.

README.md
```diff
@@ -191,6 +193,15 @@ account](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_en
 You can also enable this per-service account with the annotation
 `eks.amazonaws.com/sts-regional-endpoints` set to `"true"`.
 
+### pod-identity-webhook ConfigMap
```
nckturner, line 196: Thanks for adding this.  I think I'd like to see two things, a short example of the configmap contents, and a note/link that this was added for kops and similar use cases (I can imagine it might be a bit user unfriendly to add the configmap and update it by hand).

nckturner (commented):
Thanks for the additional documentation -- one remaining question I have is what is the behavior if the configmap is improperly formatted?  The unmarshal should fail, populateCacheFromCM should return an error, but it shouldn't affect the service account informer.  Is that correct?

README.md
olemarkus: Thought of that, but felt it may be a bit verbose. Added now.

nckturner: Thanks.  We can always move to secondary readme page to avoid verbosity but I think its good to have it documented.

nckturner (approved):

PR #139, opened 2022-01-24
add cert manager deployment

prasita123:
*Issue #, if available:*

*Description of changes:* updates to
* deploy/deployment-base.yaml
  - setting in-cluster=false
  - update volumeMounts paths
  - add cert-manager deployment
     - cert-manager.io/v1 ClusterIssuer
     - cert-manager.io/v1 Certificate
* deploy/mutatingwebhook.yaml
  - add annotations to replace caBundle

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #143, opened 2022-02-11
Fix manifests and Makefile for v0.3.0

h3poteto:
*Issue #, if available:*

This pull request is related to #139 and #134.

*Description of changes:*
- deploy
    - Remove unnecessary roles
    - Add a kustomization file
- Makefile
    - Stop to override mutating webhook configuration file
    - Stop checking and applying CSR

After v0.3.0, `--in-ncluster=true` option does not work fine. But Makefile was supposed to use CSR, so I could not install manifests using Makefile. I change manifests and Makefile to not use CSR.


deploy/auth.yaml
```diff
@@ -5,43 +5,6 @@ metadata:
...
-  namespace: default
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: pod-identity-webhook
-subjects:
-- kind: ServiceAccount
-  name: pod-identity-webhook
-  namespace: default
----
```
h3poteto, line 43: When `--in-cluster=false`, we don't need this role. Because this webhook no longer creates a secret. The secret will be created by cert-manager.

Issue #130, opened 2021-12-08
AWS_STS_REGIONAL_ENDPOINTS default to regional


philmcdo:
**What would you like to be added**:

Change the default for the webhook to inject `AWS_STS_REGIONAL_ENDPOINTS: regional` as the default instead of being the override option. Setting STS to the global endpoint can instead be the override option.

**Why is this needed**:

Defaulting AWS_STS_REGIONAL_ENDPOINTS to TRUE would follow AWS best practices for using STS: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html

> AWS recommends using Regional AWS STS endpoints instead of the global endpoint to reduce latency, build in redundancy, and increase session token validity.

mmerickel:
This should be the default for EKS. The docs state clearly:

> This is desirable in almost all cases, unless the STS regional endpoint is disabled in your account.

Shouldn't the desirable version be the default? There doesn't appear to be any support in EKS to override this default without deleting the default mutatingwebhookconfiguration and installing this config from scratch. You'd think it'd at least be something configurable as an EKS addon.

jaypipes:
This seems like a reasonable request to me. What do you think @micahhausler?

jyotimahapatra:
EKS is tracking a workitem to make the default config as regional. However, there could be customers who could have disabled regional sts endpoints and still able to use pod identity webhook today because its reaching the global endpoint. Making a sweeping retroactive change would break these customers. We're evaluating with STS to check whether any customer is impacted by this usecase. Once the evaluation is done, we can make the change on all EKS clusters.

jyotimahapatra:
EKS has completed the evaluation and is currently making changes to make the default configuration use regional sts endpoint. The changes will be rolled out for new EKS 1.16-1.21 clusters by Q1 2022. We'll update the docs once the changes are deployed.

rittneje:
@jyotimahapatra This change broke us, because it changed the pre-signed URLs generated by the AWS SDK. Docs should have been updated and customers notified well in advance, rather than finding out by surprise when we have an incident.

jyotimahapatra:
@rittneje can you help describe the failure mode?

jyotimahapatra:
@rittneje We have rolled out the docs last week around the change. We'll evaluate if we need a global reachout for the change. Let me know if you think we should take any other action from code/config perspective.

jyotimahapatra:
[...]

PR #148, opened 2022-03-15
Update default sts to regional

jyotimahapatra:
As part of https://github.com/aws/amazon-eks-pod-identity-webhook/issues/130 EKS has moved the default to regional sts. This PR makes the default flag to `true` to keep the behavior consistent across upstream and EKS.


nckturner (approved):

jyotimahapatra:
Not merging yet. There were some customer issues due to this change and we're reverting changing the default on 1.17-1.21 clusters

Issue #155, opened 2022-06-10
Can't configure MinTokenExpiration


everpeace:
**What would you like to be added**:

make `MinTokenExpiration` configurable because there is no way to configure token expiration duration less than this value

**Why is this needed**:

Although the comment of `MinTokenExpiration` says:

https://github.com/aws/amazon-eks-pod-identity-webhook/blob/0d254eee1537e0745679252ca60f020fa1a461f0/pkg/constants.go#L21-L22

k8s API says:

https://github.com/kubernetes/api/blob/1eb735bf86030410bc1d8323332500c7502dad70/core/v1/types.go#L1667-L1674

```golang
type ServiceAccountTokenProjection struct {
...
	// expirationSeconds is the requested duration of validity of the service
	// account token. As the token approaches expiration, the kubelet volume
	// plugin will proactively rotate the service account token. The kubelet will
	// start trying to rotate the token if the token is older than 80 percent of
	// its time to live or if the token is older than 24 hours.Defaults to 1 hour
	// and must be at least 10 minutes.
	//+optional
	ExpirationSeconds *int64 `json:"expirationSeconds,omitempty" protobuf:"varint,2,opt,name=expirationSeconds"`
```

Thus, I think making `MinTokenExpiration` configurable with validation that guarantees configured value must be at least 10 min as documented.

I'm very welcome to contribute to this.

PR #156, opened 2022-06-10
make MinTokenExpiration configurable by cli flag

everpeace:
*Issue #, if available:* https://github.com/aws/amazon-eks-pod-identity-webhook/issues/155

*Description of changes:*

- Changed `MinTokenExpiration` from `const` to `var`
- Introduced a cli flag `--min-token-expiration` to configure `MinTokenExpiration` with validation(it must be at least 10min).


jaypipes:
@micahhausler any reason why `MinimumTokenExpiration` was set to a constant?

@everpeace can you elaborate on the use case that a configurable minimum token expiration value is providing?

everpeace:
I just want to make k8s SA tokens refresh in roughly the same frequency as the default AWS session expiration. 

When we assume some security breach scenario, a longer expiration k8s token allows attackers more chances to impersonate. AWS sessions can revoke timely on the AWS management console, but revoking k8s SA tokens timely is relatively difficult because the cluster operator must rotate the signing key of SA tokens, especially in the on-premise cluster cases.

Thus, I think making the value configurable allows users to be able to control their security risk by themselves.

What do you think?

everpeace:
@jaypipes @micahhausler could you check my PR again??

micahhausler:
Hey sorry, I have no objection to adding this flag. I'd rather see a value plumbed down than a global variable.

Issue #78, opened 2020-09-17
IAM role alias


Angry-Potato:
**What would you like to be added**:

If role aliasing (see [IoT IAM](https://docs.aws.amazon.com/iot/latest/developerguide/security_iam_service-with-iam.html) for details on role aliases) comes to be rolled out to broader AWS, I would like to be able to use a role alias rather than a full role ARN when annotating a service account / pod with an IAM role.

**Why is this needed**:

This would be useful for reducing automation code in the following scenario.

You have multiple AWS accounts, each with kubernetes cluster(s) running for any given business reason (departments, multi-tenancy, etc.).

Using a centralised CI system, you have one application that you wish to deploy to a number of clusters in separate accounts. Each instance of this application will need to interact with the AWS account it is deployed in, and will need an AWS role in each account accordingly.

Currently the deployment pipeline for said application would look something like:

1. test / build
2. get / create / interpolate role for the application to use on deployment
3. package
4. deploy
5. test / verify / notify

This pipeline could be simplified by the requested feature if you know ahead of time that a role of a given alias will always exist in the target accounts, then there is no need to interpolate account ID.

chlunde:
Hi, we have the same issue. I'm not sure if we're looking for the exact same _solution_, because I didn't find out what IoT role aliases are based on that link :) Sorry if I'm hijacking this issue and this does not solve your problem:

I propose that either
* `role/s3-reader` (I like this best, are there any downsides?)
* `arn:aws:iam:::role/s3-reader`
* `arn:aws:iam::$AWS_ACCOUNT:role/s3-reader`

expands to `arn:aws:iam::111122223333:role/s3-reader` (assuming the EKS cluster is in account 111122223333).

```diff
diff --git a/pkg/cache/cache.go b/pkg/cache/cache.go
index b899c6a..74776b9 100644
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -22,7 +22,8 @@ import (
        "time"
 
        "github.com/aws/amazon-eks-pod-identity-webhook/pkg"
        "k8s.io/api/core/v1"
+       "github.com/aws/aws-sdk-go/aws/arn"
        "k8s.io/apimachinery/pkg/fields"
        "k8s.io/apimachinery/pkg/labels"
        utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -51,6 +52,7 @@ type serviceAccountCache struct {
        annotationPrefix   string
[...]

PR #169, opened 2023-02-01
Compose full IAM ARN from role name or alias

zewolfe:
*Issue #, if available:*
Reference: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/78

*Description of changes:*
This PR introduces the ability to create a fully formed IAM ARN if just the role alias/name is specified. For example an input of `s3-reader` will result in `arn:aws:iam::111122223333:role/s3-reader`. The account ID is fetched from ec2 Metadata


main.go
```diff
@@ -115,13 +118,27 @@ func main() {
...
+
+	sess, err := session.NewSession()
+	if err != nil {
+		klog.Fatalf("Error creating session: %v", err.Error())
+	}
+
+	metadataClient := ec2metadata.New(sess)
+	identity, err := metadataClient.GetInstanceIdentityDocument()
+	if err != nil {
+		klog.Fatalf("Error getting instance identity document: %v", err.Error())
```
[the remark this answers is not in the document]
nckturner: Can we only fail on this dependency if the feature is enabled?

pkg/cache/cache.go
```diff
@@ -142,6 +145,20 @@ func (c *serviceAccountCache) ToJSON() string {
...
 func (c *serviceAccountCache) addSA(sa *v1.ServiceAccount) {
 	arn, ok := sa.Annotations[c.annotationPrefix+"/"+pkg.RoleARNAnnotation]
+
+	if !strings.Contains(arn, "arn:") && c.composeRoleArn {
+		var accountId, partition string
+		identity := c.identity
+
+		accountId = identity.AccountID
+		if strings.Contains(identity.Region, "cn-") {
+			partition = "aws-cn"
```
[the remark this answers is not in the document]
njuettner: I added the other partitions.

main.go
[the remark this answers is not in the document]
njuettner: Ok should be done, I just wrapped a if condition around it 🙂

njuettner:
Thanks again for your input, I hope I addressed your comments. PTAL again ❤️

njuettner:
@nckturner @dims could you take a look again please?

njuettner:
@nckturner @dims I know you're super busy, I'd would really like to see this move. Is there any chance you have a couple of minutes to have a look again or you might point me to another person who can take a look, I don't want to waste your time but I'm a bit clueless here how we can move forward.

Thank you again ❤️

dims:
cc @nnmin-aws

nnmin-aws:
thanks @dims! I will review it today.

main.go
```diff
@@ -115,13 +118,27 @@ func main() {
...
+
 	saCache := cache.New(
 		*audience,
 		*annotationPrefix,
 		*regionalSTS,
+		*composeRoleArn,
 		*tokenExpiration,
 		saInformer,
 		cmInformer,
+		identity,
```
nckturner, line 144: Passing the entire instance identity document to the cache seems excessive and leaky, can we instead create a struct which includes all the data required for this feature, called something like `ComposeRoleArn`, which contains `Enabled`, `Region`, `Partition` and `AccountId`?

njuettner: Ok I changed it slightly, I hope this is better now.

njuettner:
@nckturner Addressed the feedback, thank you 👍🏻

nckturner:
@njuettner thanks for doing that, taking a look.

main.go
```diff
@@ -70,6 +73,7 @@ func main() {
 	region := flag.String("aws-default-region", "", "If set, AWS_DEFAULT_REGION and AWS_REGION will be set to this value in mutated containers")
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
 	watchConfigMap := flag.Bool("watch-config-map", false, "Enables watching serviceaccounts that are configured through the pod-identity-webhook configmap instead of using annotations")
+	composeRoleArn := flag.Bool("compose-role-arn", false, "Whether to compose a fully formed role arn when we detect one that is not fully qualified")
```
nckturner, line 76: So it has to be the role name and path, right?  Or does just the role name work?  Can you make the comment a little more specific, something like the following (but correct the verbiage if its incorrect):

```suggestion
	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
```

pkg/cache/cache.go
```diff
@@ -142,6 +151,11 @@ func (c *serviceAccountCache) ToJSON() string {
 
 func (c *serviceAccountCache) addSA(sa *v1.ServiceAccount) {
 	arn, ok := sa.Annotations[c.annotationPrefix+"/"+pkg.RoleARNAnnotation]
+
+	if !strings.Contains(arn, "arn:") && c.composeRoleArn.Enabled {
+		arn = fmt.Sprintf("arn:%s:iam::%s:role/%s", c.composeRoleArn.Partition, c.composeRoleArn.AccountID, arn)
+	}
+
```
nckturner, line 158: I guess we don't validate the `role-arn` annotation that I see.  After this if block, it would be nice if we compared the value of `arn` to an ARN regex and logged an error if it doesn't match the regex.  Just in case someone misinterprets this new parameter, or has a typo, and also just because we should have had validation in the first place.  (Unless I'm missing it, I didn't look closely).  

Here is some code and a possible regex you can use:

```suggestion
	if !strings.Contains(arn, "arn:") && c.composeRoleArn.Enabled {
		arn = fmt.Sprintf("arn:%s:iam::%s:role/%s", c.composeRoleArn.Partition, c.composeRoleArn.AccountID, arn)
	}

	matched, err := regexp.Match(`^arn:aws[a-z0-9-]*:iam::\d{12}:role\/[\w-\/.@+=,]+$`, []byte(arn))
	if err != nil {
		klog.Errorf("Regex error: %v", err)
	} else if !matched {
		klog.Warningf("arn is invalid: %s", arn)
	}
```

What do you think?

nckturner (commented):
Looks good!  Just two minor suggestions, otherwise lgtm.

njuettner:
@nckturner Awesome, thank you 👍🏻. I committed your two suggestions.

njuettner:
@nckturner if you have time again to have a final look that would be awesome 🙂

njuettner:
@nckturner any news 🙏🏻?

njuettner:
@nckturner do you mind having a look? I think we can merge this PR 🙂

nnmin-aws:
@njuettner apology for the delay as Nick is on leave. I am checking it.

nnmin-aws:
@dims Could you please kindly help merge this PR as Nick is on leave? thank you!

dims:
@nnmin-aws running the tests one last time

dims:
@zewolfe can you please rebase and fix the broken test?

njuettner:
@dims please take a look again 👍🏻

dims (approved):

dims:
thanks @njuettner

njuettner:
thank you @dims and @nckturner ❤️

dims:
and @nnmin-aws :)

calvix:
Hello  @nnmin-aws and @dims, would it be possible to create a release with this change?

calvix:
Hello  @nnmin-aws and @dims, would it be possible to create a release with this change?

nnmin-aws:
apology for your inconvenience. we will work on this and let you know when new release is out

nnmin-aws:
@calvix the new release v0.5.0 is out. thank you

Issue #165, opened 2022-12-23
Incompatible with grpc health probes


mikberg:
**What happened**:

gRPC probes were introduced in Kubernetes 1.24, adding a new field `grpc` to `Probe` (used in `readinessProbe` and `livenessProbe`). The pod identity webhook seems to be incompatible with this. Pods with a service account with the `eks.amazonaws.com/role-arn` annotation can't be created:

```
The Pod "test-grpc" is invalid: spec.containers[0].readinessProbe: Required value: must specify a handler type
```

**What you expected to happen**:

gRPC probes working, and the pod identity webhook working

**How to reproduce it (as minimally and precisely as possible)**:

Observe that this example pod can be deployed and works as expected:

```
apiVersion: v1
kind: Pod
metadata:
  name: test-grpc
spec:
  containers:
  - name: agnhost
    image: k8s.gcr.io/e2e-test-images/agnhost:2.35
    command: ["/agnhost", "grpc-health-checking"]
    ports:
    - containerPort: 5000
    - containerPort: 8080
    readinessProbe:
      grpc:
        port: 5000
```

Create a Kubernetes service account `my-sa` with the `eks.amazonaws.com/role-arn` annotation set, and try to use it in a new pod:

```
apiVersion: v1
kind: Pod
metadata:
  name: test-grpc
spec:
  serviceAccountName: my-sa
  containers:
  - name: agnhost
    image: k8s.gcr.io/e2e-test-images/agnhost:2.35
    command: ["/agnhost", "grpc-health-checking"]
    ports:
    - containerPort: 5000
    - containerPort: 8080
    readinessProbe:
      grpc:
        port: 5000
```

This error message is returned:

```
The Pod "test-grpc" is invalid: spec.containers[0].readinessProbe: Required value: must specify a handler type
```

**Anything else we need to know?**:

**Environment**:
- AWS Region: eu-north-1
- EKS Platform version: eks.3
- Kubernetes version: 1.24
- Webhook Version: ?

lareeth:
I have also just encountered this issue, and was wondering if there is a fix or a work around available yet?

lareeth:
Just to add to this, when checking the audit log, you can see the webhook patch contents, which has stripped the `grpc: {}` element

```
{
	"configuration": "pod-identity-webhook",
	"webhook": "iam-for-pods.amazonaws.com",
	"patch": [
		{
			"op": "add",
			"path": "/spec/volumes/0",
			"value": {
[...]

PR #175, opened 2023-02-13
Generate an additive JSON Patch instead of overwriting containers

iamnoah:
*Issue #, if available:* 165

*Description of changes:* 

The problem in #165 is that the code is serializing the k8s.io structs as is, with an 'add' operation that effectively replaces the Container in the PodSpec. This works until fields are added, like grpc, at which point this webhook has to be updated, or it will drop the new fields.

Instead of serializing the entire container struct, this uses jsonpatch to create a diff from the modifications, and generated a targeted patch that only adds and modifies the things it needs to. Unknown fields will not be dropped.

I've not updated the test cases as they are hardcoded to expect a specific patch (that will drop unknown fields.) TBH, that approach needs to be rethought.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/prometheus/client_golang v1.12.2
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.0
+	gomodules.xyz/jsonpatch/v2 v2.2.0
 	gopkg.in/square/go-jose.v2 v2.5.1
 	k8s.io/api v0.25.6
 	k8s.io/apimachinery v0.25.6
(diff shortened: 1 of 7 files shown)


iamnoah:
To be clear, the goal of this PR is to change the resulting patch. Here is an example diff from the unit tests:

```diff
--- want
+++ got
@@ -21,29 +21,26 @@
   },
   {
     "op": "add",
-    "path": "/spec/containers",
+    "path": "/spec/containers/0/volumeMounts",
     "value": [
       {
-        "name": "balajilovesoreos",
-        "image": "amazonlinux",
-        "env": [
+        "mountPath": "/var/run/secrets/eks.amazonaws.com/serviceaccount",
+        "name": "aws-iam-token",
+        "readOnly": true
+      }
+    ]
+  },
   {
+    "op": "add",
+    "path": "/spec/containers/0/env",
+    "value": [
+      {
             "name": "AWS_ROLE_ARN",
             "value": "arn:aws:iam::111122223333:role/s3-reader"
           },
           {
             "name": "AWS_WEB_IDENTITY_TOKEN_FILE",
             "value": "C:\\var\\run\\secrets\\eks.amazonaws.com\\serviceaccount\\token"
-          }
-        ],
-        "resources": {},
-        "volumeMounts": [
-          {
-            "name": "aws-iam-token",
-            "readOnly": true,
-            "mountPath": "/var/run/secrets/eks.amazonaws.com/serviceaccount"
-          }
-        ]
       }
     ]
   }
```

The new patch is more targeted but  it adds the same values to env and mounts. The omission of the name and image are a good thing, because that means the rest of the container spec is unchanged.

Issue #174, opened 2023-02-07
Webhook not injecting Env Vars or Volumes/VolumeMounts on initial deployment to a new cluster


ahuffman:
**What happened**:
Deploying multiple services on my cluster such as cluster-autoscaler, external-dns, ebs-csi-drivers.  On initial deployment the pods do not receive the environment vars, volumes, and volumeMounts.

When I manually delete the affected pods after automated deployment, I get everything as expected from the webhook.  

I've followed every possible AWS document on troubleshooting IRSA.  I initially thought it could be a race condition post cluster instantiation, but I tested delaying the deployments as long as 10 minutes and the results are the same.

**What you expected to happen**:
Environment vars, volumes, and volumeMounts are injected into the deployment's pod specs without need for manual deletion of the pods.

**How to reproduce it (as minimally and precisely as possible)**:
Create an EKS cluster, create an IAM OIDC provider, Create a IAM Policy, Create a Role, Attach the Policy to the Role, Create a trust relationship in the role referring to the OIDC provider and the Kubernetes service account, and finally do a helm release with appropriate values to specify the corresponding namespace, service account name, and annotations required for the roleARN to tie it all together.

**Anything else we need to know?**:
Not that I can think of, but feel free to ask for more :).

**Environment**:
- AWS Region: us-east-1 (have tested in many with same result)
- EKS Platform version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.3
- Kubernetes version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.version`): 1.24 (also tested 1.23 with same result)
- Webhook Version: ? whatever comes with EKS 1.24.8

mohammadasim:
I am experiencing the same issue. Our cluster is a Kops managed cluster. We deployed a service that had two replicas. We noticed that one of the pod was able to access the s3 bucket but the other wasn't. On investigation, the pod that was not able to access the bucket didn't have `AWS_ROLE_ARN` `AWS_WEB_IDENTITY_TOKEN_FILE` environment variables. The pod that was able to access the s3 bucket had these environment variables. I checked both pods had the same service account. When I deleted the pod with the missing environment variables, the new pod created had these two environment variables. 
My kubernetes version is `v1.22.5`.  The `amazon-eks-pod-identity-webhook` version is as follows
`Image:         amazon/amazon-eks-pod-identity-webhook:latest
[...]

PR #178, opened 2023-03-03
add check for SA to client if not found in cache

jsilverio22:
*Issue #, if available:*
#174

*Description of changes:*
Trying to solve a race condition in aws-identity pod where serviceaccount and pod get created at the same time. It uses cache to check SA which might not be ready when the pod was created. In order to solve this a call to the k8s client is used to check for a valid SA before erroring out and skipping.


jessebye:
@jsilverio22 We're encountering this issue in our clusters and wondering if this fix could get merged. Have you heard from any maintainers on this?

PR #180, opened 2023-03-07
cache: Use regexp to match wildcards stored in configmap

brenix:
*Description of changes:*

When enabling the configmap watcher, [the README provides an example configmap which uses a wildcard character for matching service accounts across multiple namespaces](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/README.md#pod-identity-webhook-configmap). However, during testing I found that this did not work due to the way that the webhook does a direct match from the corresponding service account and the cache entries. 

We desire this functionality as we have the same service account name across multiple namespaces which needs the same role ARN assigned. Since we run the same across multiple environments as well, the role ARN is different due to account IDs etc.

This PR adds functionality to support the documented feature by iterating through cache entries and performing a regex match for any wildcard characters that are defined in the configmap with the requesting namespace/service-account-name.

We have done limited testing on this and confirm that it is working as desired as well as configmap entries that do not contain wildcards.

I would welcome any suggestions or changes as I am not too familiar with the codebase (or the performance impacts of using regexp), but would love to get this merged if possible to solve our use case.

Thanks!

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -18,6 +18,7 @@ package cache
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


brenix:
After submitting this, I ran across #169 which provides a alternative approach/solution that would work for us. We can close this if the functionality covered in that can be merged. Any feedback is appreciated!

brenix:
Closing this as we solved our issue using a mutating webhook in the cluster.

PR #184, opened 2023-04-26
Enable cross-compilation in Dockerfile

chewong:
*Issue #, if available:* N/A

*Description of changes:*

Enable cross-compilation in Dockerfile. We need to add ARG TARGETOS TARGETARCH for them to be populated by buildx.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM golang:1.19 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.19 AS builder
 
 WORKDIR $GOPATH/src/github.com/aws/amazon-eks-pod-identity-webhook
 COPY . ./
-RUN GOPROXY=direct CGO_ENABLED=0 GOOS=linux go build -o /webhook -v -a -installsuffix nocgo -ldflags="-buildid='' -w -s" .
+ARG TARGETOS TARGETARCH
+RUN GOPROXY=direct CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /webhook -v -a -installsuffix nocgo -ldflags="-buildid='' -w -s" .
 
 FROM scratch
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ test:
 
 docker:
 	@echo 'Building image $(IMAGE)...'
-	docker build --no-cache -t $(IMAGE) .
+	docker buildx build --output=type=docker --platform linux/arm64 --no-cache -t $(IMAGE) .
```
nnmin-aws, line 28: does it mean `make docker` will make arm64 image always?

chewong: I can make default to `go env GOARCH` (builder's machine architecture)

dims: sounds good to me!

dims (dismissed):
/approve
/lgtm

Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM golang:1.19 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.19 AS builder
 
 WORKDIR $GOPATH/src/github.com/aws/amazon-eks-pod-identity-webhook
 COPY . ./
-RUN GOPROXY=direct CGO_ENABLED=0 GOOS=linux go build -o /webhook -v -a -installsuffix nocgo -ldflags="-buildid='' -w -s" .
+ARG TARGETOS TARGETARCH
+RUN GOPROXY=direct CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /webhook -v -a -installsuffix nocgo -ldflags="-buildid='' -w -s" .
```
nckturner, line 6: What is the difference between using nocgo and `CGO_ENABLED=0`?  Not sure if I've seen nocgo before.

chewong: I don't think it's needed anymore.

nckturner (approved):

PR #189, opened 2023-08-10
Add support for container credentials method

philljie:
*Issue #, if available:*

*Description of changes:*

Prior this change, the webhook expects the IAM Role ARN to be specified during pod admission.  The webhook mutates the pod spec by injecting `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` env vars, which will instruct the AWS SDK to get credentials via the [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc) method.

This PR introduces a new method that utilizes [AWS SDK Containers Credential Provider](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html).  This method mutates the pod spec by injecting `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` env vars, which will instruct the AWS SDK to get credentials through the specified HTTP endpoint.  

To enable this new method, user must provide a config file by setting the argument `--watch-config-file=<path>`.  Pod will use this method if its namespace & serviceAccount are listed in the config file.  The config file should be a JSON file with the following format
```
{
  "identities": [
    {
      "namespace": "foo",
      "serviceAccount": "bar"
    }
  ]
}
```

*Testing:*
- New unit tests
- Manual testing with the below

First, the webhook is configured to use the following config file
```
{
  "identities": [
    {
      "namespace": "foo",
      "serviceAccount": "bar"
    }
  ]
}
```

Second, apply the following yaml
```
apiVersion: v1
kind: Namespace
metadata:
 name: "foo"
---
apiVersion: v1
kind: ServiceAccount
metadata:
 name: "bar"
 namespace: foo
automountServiceAccountToken: false
---
apiVersion: v1
kind: Pod
metadata:
 name: test-pod-2
 namespace: foo
spec:
 serviceAccountName: bar
 containers:
 - name: test-container
   image: public.ecr.aws/eks-distro-build-tooling/builder-base:latest
   command: [ "/bin/bash", "-c", "--" ]
   args: [ "while true; do sleep 5; done" ]
---
```

Lastly, check mutated pod spec
```
spec:
  containers:
  - args:
    - while true; do sleep 5; done
    command:
    - /bin/bash
    - -c
    - --
    env:
    - name: AWS_DEFAULT_REGION
      value: us-west-2
    - name: AWS_REGION
      value: us-west-2
    - name: AWS_CONTAINER_CREDENTIALS_FULL_URI
      value: http://169.254.170.23/v1/credentials
    - name: AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE
      value: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
    image: public.ecr.aws/eks-distro-build-tooling/builder-base:latest
    imagePullPolicy: Always
    name: test-container
    ...
    volumeMounts:
    - mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
      name: aws-iam-token
      readOnly: true
  ...
  volumes:
  - name: aws-iam-token
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          audience: pods.eks.amazonaws.com
          expirationSeconds: 86400
          path: token
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -75,9 +75,9 @@ func main() {
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
 	watchConfigMap := flag.Bool("watch-config-map", false, "Enables watching serviceaccounts that are configured through the pod-identity-webhook configmap instead of using annotations")
 	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
-	watchConfigFile := flag.String("watch-config-file", "", "ac")
-	containerCredentialsAudience := flag.String("container-credentials-audience", "sts.amazonaws.com", "")
-	containerCredentialsFullUri := flag.String("container-credentials-full-uri", "", "")
+	watchConfigFile := flag.String("watch-config-file", "", "Absolute path to the config file to watch for")
+	containerCredentialsAudience := flag.String("container-credentials-audience", "pods.eks.amazonaws.com", "The audience for tokens used by the AWS Container Credentials method")
+	containerCredentialsFullUri := flag.String("container-credentials-full-uri", "http://169.254.170.23/v1/credentials", "AWS_CONTAINER_CREDENTIALS_FULL_URI will be set to this value in mutated containers")
 
 	version := flag.Bool("version", false, "Display the version and exit")
 
--- a/main.go
+++ b/main.go
@@ -21,7 +21,7 @@ import (
 	"crypto/x509/pkix"
 	goflag "flag"
 	"fmt"
-	identityconfig "github.com/aws/amazon-eks-pod-identity-webhook/pkg/config"
+	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/containercredentials"
 	"net/http"
 	"os"
 	"strings"
@@ -75,7 +75,7 @@ func main() {
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
 	watchConfigMap := flag.Bool("watch-config-map", false, "Enables watching serviceaccounts that are configured through the pod-identity-webhook configmap instead of using annotations")
 	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
-	watchConfigFile := flag.String("watch-config-file", "", "Absolute path to the config file to watch for")
+	watchContainerCredentialsConfig := flag.String("watch-container-credentials-config", "", "Absolute path to the container credential config file to watch for")
 	containerCredentialsAudience := flag.String("container-credentials-audience", "pods.eks.amazonaws.com", "The audience for tokens used by the AWS Container Credentials method")
 	containerCredentialsFullUri := flag.String("container-credentials-full-uri", "http://169.254.170.23/v1/credentials", "AWS_CONTAINER_CREDENTIALS_FULL_URI will be set to this value in mutated containers")
 
@@ -185,12 +185,12 @@ func main() {
 	saCache.Start(stop)
 	defer close(stop)
 
-	fileConfig := identityconfig.NewFileConfig(*containerCredentialsAudience, *containerCredentialsFullUri)
-	if watchConfigFile != nil && *watchConfigFile != "" {
-		klog.Infof("Watching config file %s", *watchConfigFile)
-		err = fileConfig.StartWatcher(signalHandlerCtx, *watchConfigFile)
+	containerCredentialsConfig := containercredentials.NewFileConfig(*containerCredentialsAudience, *containerCredentialsFullUri)
+	if watchContainerCredentialsConfig != nil && *watchContainerCredentialsConfig != "" {
+		klog.Infof("Watching container credentials config file %s", *watchContainerCredentialsConfig)
+		err = containerCredentialsConfig.StartWatcher(signalHandlerCtx, *watchContainerCredentialsConfig)
 		if err != nil {
-			klog.Fatalf("Error starting watcher on file %v: %v", *watchConfigFile, err.Error())
+			klog.Fatalf("Error starting watcher on file %v: %v", *watchContainerCredentialsConfig, err.Error())
 		}
 	}
 
@@ -198,7 +198,7 @@ func main() {
 		handler.WithAnnotationDomain(*annotationPrefix),
 		handler.WithMountPath(*mountPath),
 		handler.WithServiceAccountCache(saCache),
-		handler.WithConfig(fileConfig),
+		handler.WithContainerCredentialsConfig(containerCredentialsConfig),
 		handler.WithRegion(*region),
 	)
 
--- a/main.go
+++ b/main.go
@@ -186,7 +186,7 @@ func main() {
 	defer close(stop)
 
 	fileConfig := identityconfig.NewFileConfig(*containerCredentialsAudience, *containerCredentialsFullUri)
-	if *watchConfigFile != "" {
+	if watchConfigFile != nil && *watchConfigFile != "" {
 		klog.Infof("Watching config file %s", *watchConfigFile)
 		err = fileConfig.StartWatcher(signalHandlerCtx, *watchConfigFile)
 		if err != nil {
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ import (
 	"crypto/x509/pkix"
 	goflag "flag"
 	"fmt"
+	identityconfig "github.com/aws/amazon-eks-pod-identity-webhook/pkg/config"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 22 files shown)


main.go
```diff
@@ -74,6 +75,9 @@ func main() {
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
 	watchConfigMap := flag.Bool("watch-config-map", false, "Enables watching serviceaccounts that are configured through the pod-identity-webhook configmap instead of using annotations")
 	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
+	watchConfigFile := flag.String("watch-config-file", "", "Absolute path to the config file to watch for")
+	containerCredentialsAudience := flag.String("container-credentials-audience", "pods.eks.amazonaws.com", "The audience for tokens used by the AWS Container Credentials method")
+	containerCredentialsFullUri := flag.String("container-credentials-full-uri", "http://169.254.170.23/v1/credentials", "AWS_CONTAINER_CREDENTIALS_FULL_URI will be set to this value in mutated containers")
```
vpineda1996, line 80: I think we need to be a bit smart about this URI. We need to prioritize calling the agent through its IPv6 endpoint (if supported) and if not, fallback to its IPv4 addr.

pkg/config/config_test.go
```diff
@@ -0,0 +1,149 @@
+package config
```
vpineda1996, line 1: I think you need to inject the copyright notice to all new files. Right?

main.go
```diff
@@ -178,10 +185,20 @@ func main() {
 	saCache.Start(stop)
 	defer close(stop)
 
+	fileConfig := identityconfig.NewFileConfig(*containerCredentialsAudience, *containerCredentialsFullUri)
+	if *watchConfigFile != "" {
```
vpineda1996, line 189: Nit: `if watchConfigFile != nil && *watchConfigFile != ""`

pkg/config/config.go
```diff
@@ -0,0 +1,92 @@
...
+func NewFileConfig(containersCredentialsAudience, containersCredentialsFullUri string) *FileConfig {
+	return &FileConfig{
+		containerCredentialsAudience: containersCredentialsAudience,
+		containersCredentialsFullUri: containersCredentialsFullUri,
+		identityConfigObject:         nil,
+		cache:                        make(map[Identity]bool),
+	}
+}
+
+func (f *FileConfig) StartWatcher(ctx context.Context, filePath string) error {
```
vpineda1996, line 40: Can you leave some comments about what this `ctx` represent? Does it contain a deadline of how fast the `Watch` action should be or how long we will be watching the file for?

pkg/config/config.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type FileConfig struct {
+	containerCredentialsAudience string
+	containersCredentialsFullUri string
+
+	watcher              *filesystem.FileWatcher
+	identityConfigObject *IdentityConfigObject
+	cache                map[Identity]bool
+	mu                   sync.RWMutex // guards cache
```
vpineda1996, line 23: I would store this as a `ptr`, given that if somebody decides to "copy" the `FileConfig` object, the copy will retain a copy to the same cache, watcher and `identityConfigObject` but it will have an entirely different mutex associated with it.

pkg/constants.go
```diff
@@ -21,3 +21,8 @@ const (
 	// 1hr is min for kube-apiserver
 	MinTokenExpiration = int64(3600)
 )
+
+const (
+	AwsEnvVarContainerCredentialsFullUri     = "AWS_CONTAINER_CREDENTIALS_FULL_URI"
```
vpineda1996, line 26: Some comments about what this does will be great :)

pkg/constants.go
```diff
@@ -21,3 +21,8 @@ const (
 	// 1hr is min for kube-apiserver
```
vpineda1996, line 21: Why not join these two into a single `const` block?

pkg/filesystem/file_watcher.go
```diff
@@ -0,0 +1,126 @@
...
+
+// processEvent adds an item to the workqueue.
+func (f *FileWatcher) processEvent(event fsnotify.Event) {
+	if event.Name == f.path {
+		f.queue.Add(workItemKey)
+	}
+}
+
+func (f *FileWatcher) runWorker(ctx context.Context) {
+	for f.processNextWorkItem(ctx) {
```
vpineda1996, line 95: _Nit_: No need for a newline

pkg/filesystem/file_watcher.go
```diff
@@ -0,0 +1,126 @@
...
+	// Trigger initial file load
+	f.queue.Add(workItemKey)
+
+	var err error
+	f.watcher, err = fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	go wait.UntilWithContext(ctx, f.runWorker, time.Second)
```
vpineda1996, line 59: Can we give a name to our magic constant here? Also why poll every second?

pkg/filesystem/file_watcher.go
```diff
@@ -0,0 +1,126 @@
...
+		f.queue.Add(workItemKey)
+	}
+}
+
+func (f *FileWatcher) runWorker(ctx context.Context) {
+	for f.processNextWorkItem(ctx) {
+	}
+}
+
+func (f *FileWatcher) processNextWorkItem(ctx context.Context) bool {
```
vpineda1996, line 99: Can we make this more explicit? Some nice enum that says continue, stop polling?

I also wonder. What if we have an event that is always failing, when will we stop processing it? We always re-queue it. We will spend all of our "rate" buckets on a useless message until it goes away. Is that desirable?

What if we have really slow I/O s.t. it makes this function run slower than the refill rate, wouldn't that make this go into an infinite loop, just wasting resources? I am thinking of a corrupted disk. In that scenario, I/O is really slow and part of the file might be allocated to two disk sectors. If the second one is corrupted, the OS will error out and we will keep retrying.

pkg/config/config.go
philljie: I will keep it as is.  Looking at the k8s codebase and the webhook code base, most structs use a mixture of ptr & non-ptr.  For Mutex, most golang examples dont declare it as a pointer.

pkg/filesystem/file_watcher.go
philljie: Updating the naming.

For the re-queueing, i think the current implementation is okay.  The code is using a rate limited queue, which uses exponential backoff upon error.

pkg/handler/handler.go
```diff
@@ -354,6 +385,50 @@ func (m *Modifier) getPodSpecPatch(pod *corev1.Pod, roleName, audience string, r
...
+//
+// Some mutation parameters can be overridden via pod or serviceaccount
+// annotations. The serviceaccount cache already parsed the serviceaccount
+// annotations and flags such that annotations take precedence.
+// audience:        serviceaccount annotation > flag
+// regionalSTS:     serviceaccount annotation > flag
+// tokenExpiration: pod annotation > serviceaccount annotation > flag
+func (m *Modifier) buildPodPatchConfig(pod *corev1.Pod) *podPatchConfig {
+	// Container credentials method takes precedence
+	containerCredentialsPatchConfig := m.Config.Get(pod.Namespace, pod.Spec.ServiceAccountName)
```
nckturner, line 400: So the behavior is: if the service account in present in the file, use the container cred env vars, otherwise use the web identity env vars.  The user will therefore populate the dynamic config file with a service account, and all new pod launches that reference that service account will get the container credential env vars. Then if the service account is removed from the file, launched pods will get the web identity env vars again.    As I mentioned in a previous comment, since the env var type is controlled by the SA's presence in `--watch-config-file` we should rename that flag or improve the description to show that it is tied to the feature.  

This reinforces the config file being explicitly tied to the container credential behavior, so if we think non-eks users might want to use the file with web identity behavior then perhaps we shouldn't go in that direction, and might want another configuration option which determines how the config file is used. That being said, its probably better to wait for someone to ask for that behavior before implementing.

pkg/handler/handler.go
```diff
@@ -161,52 +186,61 @@ func (m *Modifier) addEnvToContainer(container *corev1.Container, tokenFilePath,
 		}
 	}
 
-	if reservedKeysDefined && regionKeyDefined && regionalStsKeyDefined {
-		klog.V(4).Infof("Container %s has necessary env variables already present",
-			container.Name)
+	if ((patchConfig.WebIdentityPatchConfig != nil && webIdentityKeysDefined) ||
+		(patchConfig.ContainerCredentialsPatchConfig != nil && containerCredentialsKeysDefined)) &&
+		regionKeyDefined && regionalStsKeyDefined {
+		klog.V(4).Infof("Container %s has necessary env variables already present", container.Name)
```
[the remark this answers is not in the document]
philljie: add log to print our the env var above

main.go
```diff
@@ -74,6 +75,9 @@ func main() {
 	regionalSTS := flag.Bool("sts-regional-endpoint", false, "Whether to inject the AWS_STS_REGIONAL_ENDPOINTS=regional env var in mutated pods. Defaults to `false`.")
 	watchConfigMap := flag.Bool("watch-config-map", false, "Enables watching serviceaccounts that are configured through the pod-identity-webhook configmap instead of using annotations")
 	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
+	watchConfigFile := flag.String("watch-config-file", "", "Absolute path to the config file to watch for")
```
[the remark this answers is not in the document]
philljie: renamed it to watch-container-credentials-config

pkg/handler/handler.go
philljie: Done. Renamed newly introduced things to container-credentials

main.go
philljie: Yes, either or.  When cluster is launch.  uri will be set to ipv6 if the cluster's IP Family is set to IPV6

pkg/handler/handler.go
```diff
@@ -133,80 +150,101 @@ func getContainersToSkip(annotationDomain string, pod *corev1.Pod) map[string]bo
...
 	}
 	stsKey := "AWS_STS_REGIONAL_ENDPOINTS"
 	for _, env := range container.Env {
-		if _, ok := reservedKeys[env.Name]; ok {
-			reservedKeysDefined = true
+		if _, ok := webIdentityKeys[env.Name]; ok {
+			klog.V(4).Infof("Web identity env variable %s is already defined in the pod spec", env)
+			webIdentityKeysDefined = true
+		}
+		if _, ok := containerCredentialsKeys[env.Name]; ok {
```
nckturner, line 178: What happens if only one of these env vars are set by the customer?  Does that result in a working pod state?  Right now, if only 1 is set, we say `containerCredentialsKeysDefined = true` and we skip patching entirely.  Wondering if it makes more sense to check for both and only skip patching if they are both set.  Might be confusing to diverge from the web identity logic here but trying to understand what UX makes the most sense.  Also not sure what the patch behavior is when only 1 is set (does the set env var get overwritten?).

nckturner (approved):

pkg/handler/handler.go
nckturner: Had a conversation off-line and am satisfied to keep this consistent with web identity. We can consider alternative behavior in the future gated by a flag or something if the use case arises

PR #196, opened 2023-10-18
Update container credentials method to use a different mount path

philljie:
*Description of changes:*
Update container credentials method to use a different mount path than the STS AssumeRoleWithWebIdentity method.  Prior this change, both methods will use the same mount path for the service account token.  However, this approach may cause confusions during migration from the AssumeRoleWithWebIdentity method to the Container credentials method.

Suppose user has the following AWS config file, which would signal SDK to use the AssumeRoleWithWebIdentity method.
```
[profile my_profile]
role_arn = arn:aws:iam::11111111:role/my-cool-role
web_identity_token_file = /var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

When the container credentials method is enabled for the service account, the audience of the token in`/var/run/secrets/eks.amazonaws.com/serviceaccount/token` will be set to `pods.eks.amazonaws.com` rather than `sts.amazonaws.com`, and thus STS will return InvalidIdentityToken.

A separate mount path is helpful because
1. Removes the invalid call to STS
2. Make it explicit that the token is not sharable between the two methods.

*Testing:*
Mutated pod spec before this change
```
spec:
  containers:
  - args:
    - while true; do sleep 5; done
    command:
    - /bin/bash
    - -c
    - --
    env:
    - name: AWS_DEFAULT_REGION
      value: us-west-2
    - name: AWS_REGION
      value: us-west-2
    - name: AWS_CONTAINER_CREDENTIALS_FULL_URI
      value: http://169.254.170.23/v1/credentials
    - name: AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE
      value: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
    image: public.ecr.aws/eks-distro-build-tooling/builder-base:latest
    imagePullPolicy: Always
    name: test-container
    ...
    volumeMounts:
    - mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
      name: aws-iam-token
      readOnly: true
  ...
  volumes:
  - name: aws-iam-token
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          audience: pods.eks.amazonaws.com
          expirationSeconds: 86400
          path: token
```

Mutated pod spec after this change
```
spec:
  containers:
  - args:
    - while true; do sleep 5; done
    command:
    - /bin/bash
    - -c
    - --
    env:
    - name: AWS_STS_REGIONAL_ENDPOINTS
      value: regional
    - name: AWS_DEFAULT_REGION
      value: us-west-2
    - name: AWS_REGION
      value: us-west-2
    - name: AWS_CONTAINER_CREDENTIALS_FULL_URI
      value: http://169.254.170.23/v1/credentials
    - name: AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE
      value: **/var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token**
    image: public.ecr.aws/eks-distro-build-tooling/builder-base:latest
    imagePullPolicy: Always
    name: test-container
...
    volumeMounts:
    - mountPath: /var/run/secrets/pods.eks.amazonaws.com/serviceaccount <----- Change #1a
      name: eks-pod-identity-token <----- Change #1b
      readOnly: true
...
  volumes:
  - name: eks-pod-identity-token <----- Change #2
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          audience: pods.eks.amazonaws.com
          expirationSeconds: 86400
          path: eks-pod-identity-token <----- Change #3
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -77,6 +77,9 @@ func main() {
 	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
 	watchContainerCredentialsConfig := flag.String("watch-container-credentials-config", "", "Absolute path to the container credential config file to watch for")
 	containerCredentialsAudience := flag.String("container-credentials-audience", "pods.eks.amazonaws.com", "The audience for tokens used by the AWS Container Credentials method")
+	containerCredentialsMountPath := flag.String("container-credentials-token-mount-path", "/var/run/secrets/pods.eks.amazonaws.com/serviceaccount", "The path to mount tokens used by the AWS Container Credentials method")
+	containerCredentialsVolumeName := flag.String("container-credentials-token-volume-name", "eks-pod-identity-token", "The name of the projected volume containing the injected service account token. This is only used by the AWS Container Credentials method")
+	containerCredentialsTokenPath := flag.String("container-credentials-token-path", "eks-pod-identity-token", "The path of the injected service account token. This is only used by the AWS Container Credentials method")
 	containerCredentialsFullUri := flag.String("container-credentials-full-uri", "http://169.254.170.23/v1/credentials", "AWS_CONTAINER_CREDENTIALS_FULL_URI will be set to this value in mutated containers")
 
 	version := flag.Bool("version", false, "Display the version and exit")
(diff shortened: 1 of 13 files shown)


main.go
```diff
@@ -77,6 +77,9 @@ func main() {
 	composeRoleArn := flag.Bool("compose-role-arn", false, "If true, then the role name and path can be used instead of the fully qualified ARN in the `role-arn` annotation.  In this case, webhook will look up the partition and account ID using instance metadata.  Defaults to `false`.")
 	watchContainerCredentialsConfig := flag.String("watch-container-credentials-config", "", "Absolute path to the container credential config file to watch for")
 	containerCredentialsAudience := flag.String("container-credentials-audience", "pods.eks.amazonaws.com", "The audience for tokens used by the AWS Container Credentials method")
+	containerCredentialsMountPath := flag.String("container-credentials-token-mount-path", "/var/run/secrets/pods.eks.amazonaws.com/serviceaccount", "The path to mount tokens used by the AWS Container Credentials method")
```
nckturner, line 80: Can we add something to the README which explains that "container credentials" refers to something specific w.r.t. `AWS_CONTAINER_CREDENTIALS_FULL_URI` and not just "getting credentials to a container", and explain any other differences there as well.

nckturner (approved):

PR #198, opened 2023-10-23
add metric to emit number of pods using sts web identity or container credentials

minj131:
*Issue #, if available:*

*Description of changes:*

Adds prometheus metric to emit number of pods using sts web identity or container credentials

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -19,12 +19,13 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
-	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/containercredentials"
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/containercredentials"
 
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg"
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache"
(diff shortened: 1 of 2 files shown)


pkg/handler/handler.go
```diff
@@ -404,6 +415,9 @@ func (m *Modifier) buildPodPatchConfig(pod *corev1.Pod) *podPatchConfig {
 	if containerCredentialsPatchConfig != nil {
 		regionalSTS, tokenExpiration := m.Cache.GetCommonConfigurations(pod.Spec.ServiceAccountName, pod.Namespace)
 		tokenExpiration, containersToSkip := m.parsePodAnnotations(pod, tokenExpiration)
+
+		webhookPodCount.WithLabelValues("container-credentials").Inc()
```
nckturner, line 419: Use underscores, see: https://prometheus.io/docs/practices/naming/

pkg/handler/handler.go
```diff
@@ -37,9 +39,18 @@ import (
...
+	prometheus.CounterOpts{
+		Name: "pod_identity_webhook_pod_count",
+		Help: "Indicator to how many pods are using sts web identity or container credentials",
+	}, []string{"Method"},
+)
+
 func init() {
 	_ = corev1.AddToScheme(runtimeScheme)
 	_ = admissionregistrationv1beta1.AddToScheme(runtimeScheme)
+	prometheus.MustRegister(webhookPodCount)
```
nckturner, line 53: We already have a place where we register metrics: https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/middleware.go

pkg/handler/handler.go
```diff
@@ -37,9 +39,18 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// This metric is used to count the number of pods that are mutated for either sts web identity or container credentials
+var webhookPodCount = prometheus.NewCounterVec(
```
nckturner, line 43: Similar comment as below, can you move this to be with the other metrics in https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/middleware.go

nckturner (approved):

PR #199, opened 2023-10-26
Don't push if there is no token or secret for dockerhub

nckturner:
*Issue #, if available:*

*Description of changes:*
Don't attempt to push to dockerhub if the token or secret variable isn't set. 


minj131 (approved):

josselin-c (approved):
> - name: Build container and push to Dockerhub registry

This could be updated as well.

Issue #215, opened 2024-03-11
Standard Label for Excluding Pod from Identity Mounting


DerekTBrown:
### What would you like to be added?
It would be useful to have a standard label that excludes pods from being processed by the identity webhook. 

**Proposed Change:** https://github.com/aws/amazon-eks-pod-identity-webhook/pull/216

### Why is this needed?

1. **Removing Cyclical Dependencies** - When installing a CNI and the Pod Identity Webhook onto our clusters, we identified a cyclical dependency; the identity webhook needs the network/CNI to be successfully operating to be able to communicate with the backend service, and the CNI needs the identity webhook to be functioning in order to schedule pods. The CNI doesn't rely on Amazon identities, so having a mechanism to skip the identity webhook for some pods (including the CNI) would remove this cyclical dependency.

2. **Performance/Reliability** - Having the ability to exclude pods that don't need Amazon identities can reduce load on the webhook and improve pod launching performance.  This also removes a critical dependency for services that need to be resilient to failure.

3. **Standardization is key to compatibility**- Having a canonical tag for skipping the identity webhook will make it easier for downstream projects to exclude critical components (such as CNIs) from the webhook, leading to a better out-of-the-box experience.

jeevanions:
Can't we simply avoid this by not using the Service Account that is enabled for Pod Identity webhook?

DerekTBrown:
> Can't we simply avoid this by not using the Service Account that is enabled for Pod Identity webhook?

I am not sure I understand your suggestion. How do Service Accounts fit into the mutating webhook flow?

fungusakafungus:
What exactly prevents CNI from coming up? This webhook sets `failurePolicy: Ignore`
https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/deploy/mutatingwebhook.yaml#L10
So I'd expect that it will not prevent other pods form starting if it is itself down.

DerekTBrown:
> What exactly prevents CNI from coming up? This webhook sets failurePolicy: Ignore

At least in our case, we rely on `kops` to install the webhook, which sets this to `Fail`:

https://github.com/kubernetes/kops/blob/5befa3fb96f35692411772152e2d1c22850dbc90/upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/k8s-1.16.yaml.template#L157

> Isn't this a `kops` issue then?

[...]

Issue #4, opened 2019-09-04
Helm Chart for Deployments


Joeskyyy:
**What would you like to be added**: Helm chart for standardised deployment

**Why is this needed**: Helm is widely used in the k8s community for deploying resources across clusters. This would allow for teams to bake existing workflows in things like terraform, ansible, etc. more easily than the current Makefile, as well as allow more customisation where needed for teams that may have different access control needs or configuration needs wrt some of the resources created for this.

---

Looking at the Makefile, I don't think this is an insurmountable task at all :D Looks like the only iffy part might be the cert generation?

If I can find the spare cycles, I might just pick this one up myself.

micahhausler:
Hey thanks, I think a Helm chart would be great, and I'd be happy to be a co-maintainer on one in https://github.com/helm/charts/. I think we'd want to keep it in the helm/charts repository so we can get all the CI/release goodness that is automatically setup there

Joeskyyy:
@micahhausler agreed on putting it in the helm repo, figured I'd file the FR here for tracking for the EKS side of the house. (:

max-rocket-internet:
> I think the name is great

🎉

> you should probably poke some org admin who can invite the listed OWNERS. I'll have @jqmichael and @nckturner comment on the chart PR with their owner/org request status

I'll try but I don't see much movement in that area these days. It seems it's always the same people (and bottleneck) for reviewing chart PRs.

hlascelles:
NB, for anyone arriving here: @max-rocket-internet has closed the [initial PR](https://github.com/helm/charts/pull/17099) opened in the main helm repo in favour of one in the aws repo: https://github.com/aws/eks-charts/pull/28 (also opened by him). Upvote it there (bearing in mind it is [stalled on](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/5) an a docker image build).

danopia:
For continuation, there's a new PR in the aws repo now that a docker image has been published: https://github.com/aws/eks-charts/pull/286

jkroepke:
In meantime, I provide a helm chart which optionally support the cert-manager or static generated tls certificates.

https://github.com/jkroepke/helm-charts/tree/main/charts/amazon-eks-pod-identity-webhook

LockedThread:
Any updates on this?

PR #216, opened 2024-03-11
feat: cannonical label for skipping pod identity webhook

DerekTBrown:
Adds a condition to the mutating webhook, preventing pods from being processed with a particular label.

See: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/215

### DIFF
--- a/deploy/mutatingwebhook.yaml
+++ b/deploy/mutatingwebhook.yaml
@@ -15,7 +15,7 @@ webhooks:
       path: "/mutate"
   objectSelector:
     matchExpressions:
-      - key: eks.amazonaws.com/skip-identity-webhook
+      - key: eks.amazonaws.com/skip-pod-identity-webhook
         operator: "DoesNotExist"
         values: []
   rules:
--- a/deploy/mutatingwebhook.yaml
+++ b/deploy/mutatingwebhook.yaml
@@ -13,6 +13,11 @@ webhooks:
       name: pod-identity-webhook
       namespace: default
       path: "/mutate"
+  objectSelector:
+    matchExpressions:
+      - key: eks.amazonaws.com/skip-identity-webhook
+        operator: "DoesNotExist"
+        values: []
   rules:
   - operations: [ "CREATE" ]
     apiGroups: [""]


deploy/mutatingwebhook.yaml
```diff
@@ -13,6 +13,11 @@ webhooks:
       name: pod-identity-webhook
       namespace: default
       path: "/mutate"
+  objectSelector:
+    matchExpressions:
+      - key: eks.amazonaws.com/skip-identity-webhook
```
prateekgogia, line 18: I am fine with this change, would love to hear @micahhausler opinion.

DerekTBrown: @micahhausler could you take a look at this?

DerekTBrown: @prateekgogia Can I have you re-review this?

DanielCKennedy: Can we change this to “eks.amazonaws.com/skip-pod-identity-webhook”. Identity webhook may be too generic if EKS adds more identity related webhooks in the future.

prateekgogia:
I looked into this last night, EKS users should be able to modify the mutating webhook in their cluster to skip certain pods (the change in PR).

Another option to achieve this is by using this existing annotation supported by webhook `eks.amazonaws.com/skip-containers`

Do any of these option help achieving the desired output?

DerekTBrown:
Hey @prateekgogia! 👋🏻 Thanks for looking at this!

This discussion is a bit nuanced- I've tried my best to capture this in writing, but we can also discuss synchronously if that would be easier.

> I looked into this last night, EKS users should be able to modify the mutating webhook in their cluster to skip certain pods (the change in PR).

There are a few reasons the PR is needed to solve this problem:

1. Kubernetes users at scale use IaC for managing Kubernetes resources. Users typically don't create these IaC manifests themselves, they instead rely on open-source Helm Charts, or addon/plugin managers (for example, we use [Kops](https://kops.sigs.k8s.io/addons/#pod-identity-webhook)).  At a minimum, applying your suggested change would require the user to own a fork of these manifests/helm charts (https://github.com/aws/amazon-eks-pod-identity-webhook/issues/4) indefinitely. In our case, it would actually require we own a fork of the Kops codebase to maintain this one-line change.

2. The goal of this PR and related issue (https://github.com/aws/amazon-eks-pod-identity-webhook/issues/215) is to create a set of canonical annotations that can be used to skip the `pod-identity-webhook` so that users don't need to figure this our manually. For example, `Cilium` should always skip the pod identity webhook, both because it doesn't need EKS identities, and because the `pod-identity-webhook` depends on the CNI being setup (thus creating a circular dependency). Therefore, the canonical annotation could be added to `Cilium` upstream repository to prevent this circular dependency for all users (instead of each user having to find their own solution). 

> Another option to achieve this is by using this existing annotation supported by webhook `eks.amazonaws.com/skip-containers`.

Unfortunately, this isn't a workable solution because of the circular dependency issue described [here](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/215#issue-2180458281). It appears `eks.amazonaws.com/skip-containers` is implemented inside the logic of the `eks-pod-identity-webhook` itself, but the issue here is that container doesn't start because it has no network connectivity. We need the ability for Kubernetes to skip calling the webhook, since it is non-responsive.

deploy/mutatingwebhook.yaml
DerekTBrown: @DanielCKennedy makes sense! Fixed!

prateekgogia (approved):
LGTM!

modulitos:
@DerekTBrown this change is now live across EKS.

Thank you for the contribution!

PR #224, opened 2024-04-18
updating k8s client version to 1.30 and go version to 1.22.2

jaidevmane:
*Issue #, if available:*

*Description of changes:*

Updating the k8s client libraries to 1.30, and updating go version to 1.22.1.


DanielCKennedy (commented):
k8s client libraries update to 1.30 looks good

jaidevmane:
Picked up latest version of go-runner from `https://gallery.ecr.aws/eks-distro/kubernetes/go-runner`

DanielCKennedy (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG golang_image=public.ecr.aws/eks-distro-build-tooling/golang:1.21
+ARG golang_image=public.ecr.aws/eks-distro-build-tooling/golang:1.22
```
xdu31, line 1: This is not necessary actually, because the ARG can be overridden

jaidevmane: Thanks! @xdu31. Should I change it back? I have changed references in all places to maintain consistency.
Would it be possible to get an approval?

kmala: I think its better to change it here instead of overriding when possible.

jaidevmane: Resolving this thread.

kmala (approved):

PR #229, opened 2024-07-02
Unify healthcheck endpoint port with application port

modulitos:
*Description of changes:*

Unify healthz and service port for both the traffic to be served on the same port. Prevents issues where pod is considered healthy but not serving any requests.

*Testing:*

```
# start server via "make clean && make local-serve", then:

❯ curl -k -i https://localhost:8443/healthz
HTTP/2 200
content-type: text/plain; charset=utf-8
content-length: 2
date: Tue, 02 Jul 2024 00:17:44 GMT

ok
```


kmala (approved):

jackchi:
What should the port be now?

The deployment I had for this is now failing. 
```yaml
readinessProbe:
            httpGet:
              path: /healthz
              port: metrics
              scheme: HTTP
```

kmala:
Until v0.5.5 release you shouldn't face issue as the healthz is responding on both the ports. If you using changes from master, then you should use the application port, not sure what is the name you had given for that

PR #231, opened 2024-07-09
cleanup old healthz endpoint

modulitos:
*Description of changes:*
This PR removes the deprecated /healthz endpoint as part of unifying the healthcheck with the application port.

This change is a followup from an earlier PR: https://github.com/aws/amazon-eks-pod-identity-webhook/pull/229

### DIFF
--- a/main.go
+++ b/main.go
@@ -226,11 +226,6 @@ func main() {
 
 	metricsMux := http.NewServeMux()
 	metricsMux.Handle("/metrics", promhttp.Handler())
-	// TODO: Remove this extra healthz endpoint once we've migrated
-	// the health check service to the application port:
-	metricsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
-	})
 
 	// Register debug endpoint only if flag is enabled
 	if *debug {


vbabenkoru:
Was there any deprecation period for this? Looks like the health check was just moved to another port in a single release, a breaking change.

kmala:
> Was there any deprecation period for this? Looks like the health check was just moved to another port in a single release, a breaking change.

release v0.5.5 https://github.com/aws/amazon-eks-pod-identity-webhook/releases/tag/v0.5.5 has change to publish healthz on both the ports and the next release will remove healthz from the metrics port.

> With the latest version, our health checks on the deployment stopped working because of this change and caused the pods to enter CrashLoopBackOff.

which version are you using because we haven't made a release yet with this change and we would be highlighting it as part of the release notes.

modulitos:
@vbabenkoru 

The latest release (v0.5.6) removes the old healthcheck endpoint, and highlights this breaking change in the the release notes: https://github.com/aws/amazon-eks-pod-identity-webhook/releases/tag/v0.5.6

So the deprecation period was from v0.5.5 to v0.5.6

Thank you for bringing this up 👍

PR #242, opened 2024-10-21
fetch SAs from apiserver

modulitos:
*Issue #, if available:*
https://github.com/aws/amazon-eks-pod-identity-webhook/issues/174

*Description of changes:*
Enhances the implementation introduced in https://github.com/aws/amazon-eks-pod-identity-webhook/pull/236 so that we can fetch missing service accounts from the apiserver. Retains the existing guarantees that we won't fetch multiple service accounts concurrently, minimizing load on the apiserver.

This feature is still in shadow mode (off by default).

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -304,12 +304,14 @@ func New(defaultAudience,
 
 	go func() {
 		for req := range saFetchRequests {
-			sa, err := fetchFromAPI(SAGetter, req)
-			if err != nil {
-				klog.Errorf("fetching SA: %s, but got error from API: %v", req.CacheKey(), err)
-				continue
-			}
-			c.addSA(sa)
+			go func() {
+				sa, err := fetchFromAPI(SAGetter, req)
+				if err != nil {
+					klog.Errorf("fetching SA: %s, but got error from API: %v", req.CacheKey(), err)
+					return
+				}
+				c.addSA(sa)
+			}()
 		}
 	}()
 
@@ -367,22 +369,8 @@ func fetchFromAPI(getter corev1.ServiceAccountsGetter, req *Request) (*v1.Servic
 	defer cancel()
 
 	klog.V(5).Infof("fetching SA: %s", req.CacheKey())
-	saList, err := getter.ServiceAccounts(req.Namespace).List(
-		ctx,
-		metav1.ListOptions{},
-	)
-	if err != nil {
-		return nil, err
-	}
 
-	// Find the ServiceAccount
-	for _, sa := range saList.Items {
-		if sa.Name == req.Name {
-			return &sa, nil
-
-		}
-	}
-	return nil, fmt.Errorf("no SA found in namespace: %s", req.CacheKey())
+	return getter.ServiceAccounts(req.Namespace).Get(ctx, req.Name, metav1.GetOptions{})
 }
 
 func (c *serviceAccountCache) populateCacheFromCM(oldCM, newCM *v1.ConfigMap) error {
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -16,19 +16,23 @@
 package cache
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


pkg/cache/cache.go
```diff
@@ -299,9 +299,20 @@ func New(defaultAudience, prefix string, defaultRegionalSTS bool, defaultTokenEx
 		defaultTokenExpiration: defaultTokenExpiration,
 		hasSynced:              hasSynced,
 		webhookUsage:           webhookUsage,
-		notificationHandlers:   map[string]chan struct{}{},
+		notifications:          newNotifications(saFetchRequests),
 	}
 
+	go func() {
+		for req := range saFetchRequests {
```
kmala, line 306: This would mean we are making only one request at a time to apiserver right?

pkg/cache/cache.go
```diff
@@ -351,6 +362,29 @@ func New(defaultAudience, prefix string, defaultRegionalSTS bool, defaultTokenEx
 	return c
 }
 
+func fetchFromAPI(getter corev1.ServiceAccountsGetter, req *Request) (*v1.ServiceAccount, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
+	defer cancel()
+
+	klog.V(5).Infof("fetching SA: %s", req.CacheKey())
+	saList, err := getter.ServiceAccounts(req.Namespace).List(
```
kmala, line 370: why List? why can't we use Get?

pkg/cache/cache.go
modulitos: Wrapped this in a goroutine - thanks for the catch 💯

pkg/cache/cache.go
modulitos: I moved this to Get

pkg/handler/handler.go
```diff
@@ -433,9 +433,10 @@ func (m *Modifier) buildPodPatchConfig(pod *corev1.Pod) *podPatchConfig {
 	}
 
 	// Use the STS WebIdentity method if set
-	request := cache.Request{Namespace: pod.Namespace, Name: pod.Spec.ServiceAccountName, RequestNotification: true}
+	gracePeriodEnabled := m.saLookupGraceTime > 0
+	request := cache.Request{Namespace: pod.Namespace, Name: pod.Spec.ServiceAccountName, RequestNotification: gracePeriodEnabled}
```
modulitos, line 437: I added this change to toggle `RequestNotification` only when the grace period is enabled.

Previously it was basically a no-op when the feature is disabled and `RequestNotification` is true, but now we're fetching from the API server when it's true. So we only want it set when the feature is enabled.

pkg/cache/notifications.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (n *notifications) create(req Request) <-chan struct{} {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	notifier, found := n.handlers[req.CacheKey()]
+	if !found {
+		notifier = make(chan struct{})
+		n.handlers[req.CacheKey()] = notifier
+		n.fetchRequests <- &req
```
haoranleo, line 30: We control the APIServer request rate through the size of the channel but it has two downsides:
1. The APIServer request is actually not rate limited. There is no request handling rate limiter in the channel consumption and it is possible that there could be > 100 requests (for different `namespace/name`) sent in the same second given the channel consumer initiates a new go routine to submit a request to APIServer
2. If due to some reasons the channel consumer dies or the queue is full, the `create` function will hang until the channel has some capacity. It could unexpectedly delay pod creation for arbitrary time.

A better choice could be use a larger channel size to minimize the chance of channel write blocking, and implement a more robust channel consumer which limit the consumption rate. In case of extremely high volumes of requests queued in the channel and the API requests could not be sent in time, the result would be either be the cache is synced before grace period and pod is mutated, or cache is not synced and the pod is not mutated. But no prolonged delay to pod creation or excessive requests to the APIServer

haoranleo:
Moved the changes to https://github.com/aws/amazon-eks-pod-identity-webhook/pull/252

PR #260, opened 2025-02-28
Add jitter to Default token expiration

notSoWiseOldMan:
*Issue #, if available:*
N/A

*Description of changes:*
Adds a jitter to service account tokens when using the default expiration

Additionally to handle test cases where the randInt function is called I've added mocks and create a mockery config to the root of the repo to help with generating mock interfaces (mockery is recommended by http://github.com/stretchr/testify)


pkg/constants.go
```diff
@@ -15,9 +15,10 @@ permissions and limitations under the License.
 package pkg
 
 const (
-	// Default token expiration in seconds if none is defined,
-	// which is 24hrs as that is max for EKS
-	DefaultTokenExpiration = int64(86400)
+	// 24hrs as that is max for EKS
+	MaxTokenExpiration = int64(86400)
+	// Default token expiration in seconds if none is defined, 22hrs
+	DefaultTokenExpiration = int64(79200)
```
xdu31, line 21: The `DefaultTokenExpiration` is used by both IRSAv1 and IRSAv2 (Pod Identity):
1. In IRSAv1, it has one informer cache
2. In IRSAv2, we are only using from `GetCommonConfigurations`, which is being added from onboarding to IRSAv2 https://github.com/aws/amazon-eks-pod-identity-webhook/pull/189

I believe we can add the randomization inside the `GetCommonConfigurations` function and not change the `DefaultTokenExpiration` for IRSAv1 use

pkg/handler/handler.go
```diff
@@ -479,6 +488,26 @@ func (m *Modifier) buildPodPatchConfig(pod *corev1.Pod) *podPatchConfig {
 	return nil
 }
 
+func addJitter(val int64, jitterPercent int64, min int64, max int64) (int64, error) {
```
xdu31, line 491: I think this can be just simplified to a randomization between min and max, we don't need the val. And jitterPercentage and (min, max) is kinda similar, just to add randomization

pkg/constants.go
notSoWiseOldMan: I think I have updated GetCommonConfigurations the way you are describing in the most recent commit now. Let me know if I missed somethign

pkg/handler/handler.go
```diff
@@ -459,6 +467,7 @@ func (m *Modifier) buildPodPatchConfig(pod *corev1.Pod) *podPatchConfig {
 	klog.V(5).Infof("Value of roleArn after after cache retrieval for service account %s: %s", request.CacheKey(), response.RoleARN)
 	if response.RoleARN != "" {
 		tokenExpiration, containersToSkip := m.parsePodAnnotations(pod, response.TokenExpiration)
+		tokenExpiration = m.addJitterToDefaultToken(tokenExpiration)
```
xdu31, line 470: We don't require this, this is IRSAv1 path

xdu31 (approved):

Issue #256, opened 2025-02-03
v0.6.2 of Pod Identity Webhook not published to Docker Hub


jandersen-plaid:
**What happened**: Version v0.6.2 is marked as [released on Github](https://github.com/aws/amazon-eks-pod-identity-webhook/releases/tag/v0.6.2), but is [not published to Docker Hub](https://hub.docker.com/r/amazon/amazon-eks-pod-identity-webhook/tags?name=v0.6.2)

**What you expected to happen**: I expected the version to be published to docker hub along with the other versions close to the release date of v0.6.2.

**How to reproduce it (as minimally and precisely as possible)**:

```
docker pull docker pull amazon/amazon-eks-pod-identity-webhook:v0.6.2
```

**Anything else we need to know?**: The improvements in v0.6.2 look great, so we are looking forward to integrating them!

**Environment**: N/A
- AWS Region:
- EKS Platform version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.platformVersion`):
- Kubernetes version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.version`):
- Webhook Version:

yue9944882:
@jandersen-plaid Our dockerhub credential is expired and we're still working on refreshing it atm

jandersen-plaid:
Good to know! Thanks for the response.

I will just leave this open as a tracking issue so that more people don't start filing issues for something that is already being worked on.

tmehlinger:
> Our dockerhub credential is expired and we're still working on refreshing it atm

How about publishing it to ECR Public?

seanschneeweiss:
All releases >=v0.6.3 are available on dockerhub. So probably this issue can be closed or someone would need to manually push v0.6.2 to fix.

PR #268, opened 2025-05-14
Updating actions to publish to ecr public

notSoWiseOldMan:
*Issue #, if available:*
https://github.com/aws/amazon-eks-pod-identity-webhook/issues/256

*Description of changes:*
This change updates the actions pipeline to publish to the ECR public. Currently the repo alias is still being reviewed but we can start publishing to ECR before the custom alias is applied. Images can be viewed [here.](https://gallery.ecr.aws/eks/amazon-eks-pod-identity-webhook)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,8 +6,20 @@ jobs:
     # this is to prevent the job to run at forked projects
     if: github.repository == 'aws/amazon-eks-pod-identity-webhook'
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
     - uses: actions/checkout@v3
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: arn:aws:iam::719983614556:role/github-actions-amazon-eks-pod-identity-webhook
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -16,26 +28,17 @@ jobs:
       with:
         buildx-version: latest
         qemu-version: latest
+    
     - name: Build container and push to Dockerhub registry
       run: |
         BRANCH=$(echo $GITHUB_REF | cut -d'/' -f3)
         SHORT_SHA=$(echo $GITHUB_SHA | cut -c -7)
-        REPO=amazon/amazon-eks-pod-identity-webhook
+        REPO=public.ecr.aws/c5c3y3r8/amazon-eks-pod-identity-webhook
```
kmala, line 36: why `c5c3y3r8` ? can it be something more meaningful like `eks` ?

notSoWiseOldMan: We don't get to choose this but we can set an alias. I've already requested eks.

notSoWiseOldMan: Updated

kmala (approved):

PR #299, opened 2026-03-20
updating go to v1.26

notSoWiseOldMan:
*Issue #, if available:*

*Description of changes:*
Updating Go version to resolve CVEs


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/amazon-eks-pod-identity-webhook
 
-go 1.25.7
+go 1.26
```
adp-825, line 3: just thinking, if there are any concerns in using 1.26 instead of 1.26.1

notSoWiseOldMan: I don't think so. This would help us get bug and security fixes on every build instead of just when CVEs are released

adp-825 (approved):

mengqiy (approved):

PR #301, opened 2026-04-01
feat(cache): reduce memory usage by stripping ServiceAccount objects

Swizzmaster:
*Issue #, if available:* N/A

*Description of changes:*

## Problem

The pod-identity-webhook uses a Kubernetes informer to watch all
`ServiceAccount` objects in the cluster. The informer caches the full
decoded `ServiceAccount` object for every SA, regardless of whether the
webhook actually reads any of those fields. On clusters with a large
number of ServiceAccounts (especially clusters with leaked or
auto-generated SAs from job-runner controllers), this causes the webhook
to consume hundreds of megabytes of memory, sometimes exceeding the 
container limit and OOM-killing the pod.

The webhook only reads three pieces of information from a cached SA:

1. `metadata.name`
2. `metadata.namespace`
3. Annotations beginning with the configured webhook annotation prefix
   (`eks.amazonaws.com/` by default), used to look up the IAM role and
   audience for IRSA.

Everything else the apiserver returns — `secrets`, `imagePullSecrets`,
`labels`, `managedFields`, `ownerReferences`, `finalizers`, the
`last-applied-configuration` annotation, etc. — is decoded and held in
memory on every list/watch event but never referenced.

## What this PR does

Adds a `cache.TransformFunc` to the ServiceAccount informer that
rewrites each cached object to keep only the fields the webhook needs:

- `metadata.name`
- `metadata.namespace`
- `metadata.resourceVersion` (preserved so reflector resyncs continue to
  work correctly)
- `metadata.annotations`, filtered to keys with the configured
  `--annotation-prefix` value (default `eks.amazonaws.com/`)

Every other field is dropped before the object enters the indexer.

The transform runs once per object, on the reflector goroutine, before
the object is stored in the cache. Subsequent reads from the cache see
only the stripped form, which is also what `cache.SharedIndexInformer`
hands to event handlers. Because the input object is replaced rather
than mutated in place, this is safe with respect to other informer
consumers and the upstream `client-go` cache contract.

A small `filterAnnotations` helper is added so the prefix filter is
configurable — this avoids hard-coding `eks.amazonaws.com/` and keeps
the webhook compatible with users who run with a custom
`--annotation-prefix`.

### Functional

Verified webhook behavior is unchanged:

- A pod using a ServiceAccount annotated with
  `eks.amazonaws.com/role-arn` is mutated correctly: `AWS_ROLE_ARN` and
  `AWS_WEB_IDENTITY_TOKEN_FILE` env vars are injected, and the
  projected `aws-iam-token` volume is mounted.
- A pod using a ServiceAccount with no relevant annotations is **not**
  mutated.
- A pod using a ServiceAccount with annotations that don't match the
  configured prefix is **not** mutated.

### Memory

Reproduced the shape of an OOMing cluster on a test cluster:
50,000 ServiceAccounts modeled on a real workload-runner leak pattern
(empty `secrets` / `imagePullSecrets`, ~1 KB of labels, ~1.4 KB
`last-applied-configuration` annotation, `automountServiceAccountToken:
false`, `ownerReferences` to a CRD-owned parent).

Same webhook pod, same SAs, same kubelet, same node. Sampled 5× at 30 s
intervals after a forced GC, took the median.

| metric                          | mainline | this PR | delta   |
| ------------------------------- | -------- | ------- | ------- |
| `go_memstats_heap_inuse_bytes`  | 485 MB   | 84 MB   | **−83%** |
| `go_memstats_heap_alloc_bytes`  | 466 MB   | 68 MB   | **−85%** |
| `go_memstats_heap_objects`      | 5.01 M   | 768 K   | **−85%** |
| `process_resident_memory_bytes` | 580 MB   | 150 MB  | **−74%** |

The `−85%` drop in `heap_objects` reflects that the customer's per-SA
overhead was dominated by `managedFields` and label/annotation maps —
all of which the webhook never read but were decoded into hundreds of
Go heap objects per SA by the protobuf deserializer.

### Unit tests

Existing unit tests pass. Added coverage for the transform itself:

- Known and unknown fields on input → only allowed fields on output.
- Annotation prefix filter retains matching keys, drops everything else.
- Nil-annotation case.
- Non-`*ServiceAccount` input is returned unchanged (informer transform
  contract).

## Compatibility

- No changes to webhook flags or default behavior.
- No changes to the mutation logic or admission response format.
- Cache key format is unchanged.
- Reflector / informer resync semantics are unchanged
  (`resourceVersion` is preserved on stripped objects).
- Note for future maintainers: if the webhook ever needs to read a new
  ServiceAccount field, the transform must be updated to retain it.
  This is documented in the comment block on `stripDownServiceAccount`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,8 @@ import (
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	flag "github.com/spf13/pflag"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	v1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
@@ -136,6 +138,11 @@ func main() {
 	}
 
 	saInformer := informerFactory.Core().V1().ServiceAccounts()
+
+	// Strip ServiceAccount objects down to only the fields the cache needs
+	// (Name, Namespace, Annotations). This significantly reduces memory usage
+	// of the informer store on large clusters.
+	saInformer.Informer().SetTransform(stripDownServiceAccount)
 
 	*tokenExpiration = pkg.ValidateMinTokenExpiration(*tokenExpiration)
 
@@ -371,4 +378,31 @@ func main() {
 	} else {
 		klog.Infof("Metrics server shutdown gracefully")
 	}
+}
+
+// stripDownServiceAccount is a transform function that strips down ServiceAccount
+// objects to reduce memory usage of the informer store on large clusters.
+// See details in [stripDownServiceAccountObject].
+func stripDownServiceAccount(obj interface{}) (interface{}, error) {
+	if sa, ok := obj.(*corev1.ServiceAccount); ok {
+		return stripDownServiceAccountObject(sa), nil
+	}
+	return obj, nil
+}
+
+// stripDownServiceAccountObject strips a ServiceAccount down to only the fields
+// the cache needs (Name, Namespace, Annotations, ResourceVersion).
+// NOTE: if the webhook needs to refer to more SA fields in the future, those
+// fields need to be added here.
+func stripDownServiceAccountObject(sa *corev1.ServiceAccount) *corev1.ServiceAccount {
+	sa.ObjectMeta = metav1.ObjectMeta{
+		Name:            sa.Name,
+		Namespace:       sa.Namespace,
+		Annotations:     sa.Annotations,
+		ResourceVersion: sa.ResourceVersion,
+	}
+	sa.Secrets = nil
+	sa.ImagePullSecrets = nil
+	sa.AutomountServiceAccountToken = nil
+	return sa
 }
--- a/main.go
+++ b/main.go
@@ -38,6 +38,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	flag "github.com/spf13/pflag"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	v1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/cache/cache.go
```diff
@@ -213,6 +213,17 @@ func (c *serviceAccountCache) ToJSON() string {
 }
 
 func (c *serviceAccountCache) addSA(sa *v1.ServiceAccount) {
+	// Only cache SAs that have at least one pod-identity-webhook annotation.
+	// This avoids storing every SA in the cluster and prevents OOM in large clusters.
+	if !c.hasRelevantAnnotations(sa) {
```
kmala: The issue of OOM would because of the informer cache right and not this cache as this just saves few high level details like name, namesapce

Swizzmaster: 1000%,this would fix nothing, I realized that as I was about to start testing. I’m testing a change right now using a transformer on the informer cache to strip out the unneeded information. Sorry should have left a comment on the PR.

main.go
```diff
@@ -372,3 +379,30 @@ func main() {
...
+		return stripDownServiceAccountObject(sa), nil
+	}
+	return obj, nil
+}
+
+// stripDownServiceAccountObject strips a ServiceAccount down to only the fields
+// the cache needs (Name, Namespace, Annotations, ResourceVersion).
+// NOTE: if the webhook needs to refer to more SA fields in the future, those
+// fields need to be added here.
+func stripDownServiceAccountObject(sa *corev1.ServiceAccount) *corev1.ServiceAccount {
```
yue9944882, line 397: do we have data supporting how much % memory reduction we can get from a real cluster?

Swizzmaster: Added Data to the PR details

main.go
```diff
@@ -372,3 +381,50 @@ func main() {
...
+// finalizers, UID, generation, secrets, imagePullSecrets, and unrelated
+// annotations. A fresh object is returned so the original (along with any
+// heavy fields like managedFields) can be garbage collected immediately.
+//
+// NOTE: if the webhook needs to refer to more SA fields in the future, those
+// fields need to be added here.
+func stripDownServiceAccount(annotationPrefix string) func(obj interface{}) (interface{}, error) {
+	prefix := annotationPrefix + "/"
+	return func(obj interface{}) (interface{}, error) {
+		sa, ok := obj.(*corev1.ServiceAccount)
```
mengqiy: Since we only need to access metadata fields. We can avoid type casting here by using `k8s.io/apimachinery/pkg/api/meta.Accessor`.

Swizzmaster: Good Callout, Thank you! Addressed.

mengqiy (approved):
LGTM

PR #302, opened 2026-04-08
feat: Enable use with third party providers

fitbeard:
*Issue #, if available:*

*Description of changes:*

Add new configuration flag `endpoint-url` to allow override `AWS_ENDPOINT_URL`. This allows use of this webhook with STS compatible solutions like Minio or RadosGW with non default endpoint url.

Add new configuration flag `allow-non-aws-account-id`. Relax role ARN account ID validation warnings for non-AWS IAM-compatible providers. This allows use of empty (Minio, RadosGW), 20-char (RadosGW account) or 12-digit (default) Account ID in role ARN validator.


kmala:
I don't see a need to add support for third party providers as this webhook is specifically for kubernetes clusters on AWS.

Issue #225, opened 2024-04-18
Get error during creation of pod with service-account that has eks.amazonaws.com/role-arn annotation


cynepco3hahue:
**What happened**:

One of our pods should assume a specific AWS role and we are configuring it via service account annotation. After eks cluster upgrade the mutation of the pod stopped working and it had only a default node group role.
I changed the `mutatingwebhookconfigurations` policy from Ignore to Fail for `pod-identity-webhook` under the cluster and after that pod failed to start with the error 
```
Error creating: Internal error occurred: failed calling webhook "iam-for-pods.amazonaws.com": received invalid webhook response: expected webhook response of admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind=
```
After some additional investigation, I found that the problem solved once I change
```
- admissionReviewVersions:
  - v1
```
to 
```
- admissionReviewVersions:
  - v1beta1
```
IIUC the problem that mutation webhook handler is using deprecated version of admission API `v1beta1` when it should use `v1`

**What you expected to happen**:
The mutation of the pod should work without any errors.

**How to reproduce it (as minimally and precisely as possible)**:
1. Get the EKS cluster
2. Try to configure the pod with some role, you can follow up https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

**Anything else we need to know?**:

**Environment**:
- AWS Region: us-east-1
- EKS Platform version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.platformVersion`): `eks.15`
- Kubernetes version (if using EKS, run `aws eks describe-cluster --name <name> --query cluster.version`): 1.27
- Webhook Version: Unsure how can I check it

OscarPoels:
Got the exact same problem on a EKS cluster upgrade, really don't know where it can come from as it is supposed to support both API, also when using `- admissionReviewVersions: ["v1","v1beta1"]` it does not work, only 
```
- admissionReviewVersions:
  - v1beta1
```

works

Andrei-Vasil:
Same issue here, do you have any follow up on this?

dom-colangelo:
I hit this same issue on a non-EKS cluster and traced it to the same root cause. The handler builds the response `AdmissionReview` without copying `TypeMeta` from the request, so the response serializes with empty `apiVersion`/`kind` — which Kubernetes rejects.

[...]

PR #304, opened 2026-04-19
fix: echo request TypeMeta in admission response to comply with Kubernetes webhook spec

dom-colangelo:
## Problem

The Kubernetes webhook specification requires that a webhook response include the same `apiVersion` and `kind` as the request. The handler was building the response `AdmissionReview` struct without copying TypeMeta from the request, so every response serialized as:

```json
{"response": {"uid": "...", "allowed": true, "patch": "..."}}
```

API servers enforce this strictly:

```
received invalid webhook response: expected webhook response of admission.k8s.io/v1,
Kind=AdmissionReview, got /, Kind=
```

This is a known issue — see #225, where an EKS cluster upgraded to Kubernetes 1.27 with `admissionReviewVersions: v1` began rejecting webhook responses and injection stopped working. The only workaround was to downgrade to `v1beta1`. The root cause is the same: when the API server negotiates v1, the response comes back with an empty `apiVersion`/`kind` and is rejected.

A related issue: `deserializer.Decode` used a codec scheme that registered only `corev1` and `admissionregistrationv1beta1` — not the admission API types. Any cluster sending `admission.k8s.io/v1` requests would fail to decode before mutation ran.

## Fix

**`pkg/handler/handler.go`**

1. Copy `ar.TypeMeta` into the response struct before marshalling — the response echoes back whatever `apiVersion`/`kind` the request carried, as the spec requires.
2. Replace `deserializer.Decode` with `json.Unmarshal`. `json.Unmarshal` decodes both `v1` and `v1beta1` payloads correctly without a registered scheme. The old codec scheme did not include the admission API types, so it was incapable of decoding v1 requests.
3. Remove the now-unused `runtimeScheme`, `codecs`, `deserializer` package vars and `init()`.

**`pkg/handler/handler_test.go`**

1. Update two test cases for the new `json.Unmarshal` error strings.
2. Add regression test `ValidRequestV1TypeMetaEchoed`: sends an `AdmissionReview` with `apiVersion: admission.k8s.io/v1` and asserts the response echoes it back.

## Testing

```
go test ./pkg/handler/...
```

All tests pass.

## Related

- #225 — EKS 1.27 upgrade breaks injection when `admissionReviewVersions: v1`; workaround was to force `v1beta1`

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -32,23 +32,9 @@ import (
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg"
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache"
 	"k8s.io/api/admission/v1beta1"
-	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/klog/v2"
-)
-
-func init() {
-	_ = corev1.AddToScheme(runtimeScheme)
-	_ = admissionregistrationv1beta1.AddToScheme(runtimeScheme)
-}
-
-var (
-	runtimeScheme = runtime.NewScheme()
-	codecs        = serializer.NewCodecFactory(runtimeScheme)
-	deserializer  = codecs.UniversalDeserializer()
 )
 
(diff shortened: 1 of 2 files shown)


dmetzler:
I independently hit the same issue on k3s v1.31 and arrived at a similar conclusion. Your diagnosis is spot-on — the missing TypeMeta in the response is the root cause, and EKS masking it by tolerating empty TypeMeta explains why it went unnoticed upstream.

I submitted a PR that builds on your approach (echoing `ar.TypeMeta` in the response) but also migrates the handler and tests from v1beta1 to v1 types entirely, since v1beta1 was removed in Kubernetes 1.25. See #310.

Happy to consolidate if needed!

Issue #205, opened 2024-01-16
The admissionregistration.k8s.io/v1beta1 API version of MutatingWebhookConfiguration is no longer served as of v1.22.


jackzhao0711:
Current EKS version：1.21
Purpose: Upgrade version to 1.22
EKS upgrade insights shows that there are deprecated api versions：
![image](https://github.com/aws/amazon-eks-pod-identity-webhook/assets/109124236/27114e9c-f429-4172-ad05-b3c978bf666c)

Check  cluster：
`kubectl get mutatingwebhookconfigurations.admissionregistration.k8s.io -A`
![image](https://github.com/aws/amazon-eks-pod-identity-webhook/assets/109124236/dedf2d51-38d8-4450-ab71-9e59d5288de2)

Check details：
`kubectl get MutatingWebhookConfiguration.v1beta1.admissionregistration.k8s.io "0500-amazon-eks-fargate-mutation.amazonaws.com" -o yaml | grep apiVersion`

KevinHuang40856:
for your first question, the reason the apiVersion in second column is from the rules config to lists one or more API versions to match.  if you view the full yaml output, you will find it
refer https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-rules

KevinHuang40856:
for second question, there is no such automatic upgrade. firstly all existing persisted objects are accessible via the new APIs, however as the admissionregistration.k8s.io/v1beta1 API version of MutatingWebhookConfiguration is no longer served as of v1.22 also the below changes in admissionregistration.k8s.io/v1 , you need to update YAML spec files to the new API version and change accordingly. eg you may have to add webhooks[*].admissionReviewVersions 

webhooks[*].failurePolicy default changed from Ignore to Fail for v1
webhooks[*].matchPolicy default changed from Exact to Equivalent for v1
webhooks[*].timeoutSeconds default changed from 30s to 10s for v1
webhooks[*].sideEffects default value is removed, and the field made required, and only None and NoneOnDryRun are permitted for v1
webhooks[*].admissionReviewVersions default value is removed and the field made required for v1 (supported versions for AdmissionReview are v1 and v1beta1)
webhooks[*].name must be unique in the list for objects created via admissionregistration.k8s.io/v1

jackzhao0711:
@KevinHuang40856 Thank you for your reply, since my cluster is created through CDK and these apis are automatically generated, how should I find and update these values?

[...]

Issue #132, opened 2021-12-21
admissionregistration.k8s.io/v1beta1 was deprecated in 1.22


ss098:
**What happened**:

`admissionregistration.k8s.io/v1beta1` api removed in kubernetes 1.22.

**What you expected to happen**:

[https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/deploy/mutatingwebhook.yaml#L1](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/deploy/mutatingwebhook.yaml#L1)

Migrate admissionregistration from `v1beta1` to `v1`.

h3poteto:
I think we need to merge #115 at first.

LauraCameran:
I think this issue can be closed since commit https://github.com/aws/amazon-eks-pod-identity-webhook/commit/d3d1b4cbc58e7ad25155e73778c8f80ffc6668c5

PR #310, opened 2026-08-07
fix: migrate admission API from v1beta1 to v1 and echo TypeMeta in response

dmetzler:
### Problem

The webhook handler uses admission/v1beta1 types and returns an AdmissionReview response without TypeMeta. On non-EKS clusters, the API server rejects the response:

```
received invalid webhook response: expected webhook response of
admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind=
```

EKS patches its API server to tolerate the missing TypeMeta, which is why this has gone unnoticed upstream. All self-managed deployments (k3s, kOps, kubeadm, etc.) are affected.

This has been reported multiple times: #225, #205, #132. A similar fix was proposed in #304 but never merged.

### Fix

Migrate handler and tests from admission/v1beta1 to admission/v1
Echo the request TypeMeta into the response (handles both API versions transparently, following the approach from #304)
Add regression test verifying TypeMeta is properly echoed

###Tested on

k3s v1.31 — pods are correctly mutated after the fix.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -31,8 +31,8 @@ import (
 
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg"
 	"github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache"
-	"k8s.io/api/admission/v1beta1"
-	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	admissionv1 "k8s.io/api/admission/v1"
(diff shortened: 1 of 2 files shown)


mengqiy (approved):
v1 for admission and admissionregistration groups have been available since 1.16.

mengqiy:
Prefer this PR over https://github.com/aws/amazon-eks-pod-identity-webhook/pull/304 since this one doesn't touch the deserializer.

Issue #315, opened 2026-09-08
Name-only check for existing `aws-iam-token` volume silently disables token injection but still patches env/volumeMounts into every container


KR-Ravindra:
**What happened**

If a pod already contains a volume whose *name* is `aws-iam-token` (the webhook's default `volName`), the webhook skips injecting its projected service-account-token volume, but still:

- adds `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token` to every non-skipped container, and
- adds a `volumeMount{name: aws-iam-token, mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount}` to every non-skipped container that does not already have a mount with that *name*.

Those containers now mount the *user's* volume (hostPath / emptyDir / secret / a projected token with a different `path` or `audience`) at the path the SDK is told to read the web-identity token from. The result is that the AWS SDK's web-identity provider fails (`ExpiredToken`, `InvalidIdentityToken`, or "open .../token: no such file") and falls through to the node instance role, producing `AccessDenied` from the node role, with nothing in the webhook logs except `Pod was mutated`.

This is most visible with tools that inject their own sidecars into user pods (Argo Workflows `init`/`wait`, Istio, Vault agent, log shippers, ...). The user's main container is typically fine because it already declares a mount named `aws-iam-token` and often its own credential env, so the webhook leaves it alone; only the injected sidecars, which the user never wrote, are broken. Debugging is hard because the pod looks correctly mutated: env vars and mounts are present, only the volume behind them is the wrong one.

**Code**

`pkg/handler/handler.go`:

- `getPodSpecPatch`: `// skip adding volume if it already exists` compares `vol.Name == patchConfig.VolumeName` only; the volume source is not inspected.
- `addEnvToContainer`: env is injected unless already set; the mount is appended unless a mount with the same *name* exists. Neither step is conditioned on whether the webhook actually owns the volume.
- `pkg/handler/testdata/rawPodHasVolumeMount.pod.yaml` encodes exactly this: a pod with a **hostPath** volume named `aws-iam-token` is expected to receive `AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token` while the projected volume is not added. That token file can never exist there.

[...]

PR #316, opened 2026-09-08
fix: leave pods with a foreign aws-iam-token volume unmodified

KR-Ravindra:
*Issue #, if available:* #315

*Description of changes:*

## Problem

If a pod already contains a volume *named* `aws-iam-token` (the webhook's `volName`; `con-creds-volume-name`-style names in EKS Pod Identity mode behave the same), the webhook skips adding its projected service-account-token volume but still injects `AWS_ROLE_ARN` / `AWS_WEB_IDENTITY_TOKEN_FILE` (or the container-credentials equivalents) and a `volumeMounts` entry named `aws-iam-token` into every non-skipped container. Those containers end up mounting the user's hostPath / emptyDir / secret / differently-scoped projected token at the path the SDK is told to read the web-identity token from, so credential resolution fails and falls through to the node role. This mostly hits sidecars injected by other controllers (Argo Workflows, Istio, Vault agent, ...) that the user never wrote, and the only webhook log line is `Pod was mutated`.

## Root cause

- `pkg/handler/handler.go` `getPodSpecPatch` (master line 386, `// skip adding volume if it already exists`) compares `vol.Name == patchConfig.VolumeName` only; the volume source is never inspected.
- `pkg/handler/handler.go` `addEnvToContainer` (lines 253-283) injects the env vars and appends the `aws-iam-token` mount whenever a container lacks them, independently of whether the webhook actually owns the volume behind that name.
- `pkg/handler/testdata/rawPodHasVolumeMount.pod.yaml` and `pkg/handler/testdata/containercredentials/rawPodWithConflictingVolumeMount.pod.yaml` encode this half-mutation as the expected result: a **hostPath** volume with the reserved name, and the expected patch still sets `AWS_WEB_IDENTITY_TOKEN_FILE` / `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` to a file that can never exist on that volume.

## Fix

- New `findTokenVolume` helper: a volume with the reserved name counts as *already injected* only if it is a projected volume with a `serviceAccountToken` source whose `audience` and `path` match the webhook's config. That keeps today's idempotent behaviour for `reinvocationPolicy: IfNeeded` / re-admission (volume left alone, env and mounts still ensured on containers that lack them).
- Any other volume source with that name is treated as foreign: `getPodSpecPatch` returns no patch, `MutatePod` answers `Allowed: true` without a patch (same shape as the existing "service account not annotated" path), and a `klog.Warning` at default verbosity names the pod, namespace and the volume source type, e.g. `Pod was not mutated. Reason: volume "aws-iam-token" is reserved for the projected service account token, but the pod declares a HostPath volume with that name. Pod=..., ServiceAccount=..., Namespace=...`.
- The two fixtures above now expect no patch; their previous expectation was the bug.
- README: the example volume was called `aws-token`, which does not match the code; corrected to `aws-iam-token` and added a sentence documenting that the name is reserved and what happens on a collision.

**Question for maintainers:** I went with *skip mutation + warn* because it preserves the pod's pre-webhook behaviour and cannot take down existing workloads on upgrade. The alternative is to *deny admission* with a message such as `volume "aws-iam-token" is reserved by the pod identity webhook; rename the volume`, which is louder but breaks pods that currently start (even if with the wrong credentials). Happy to switch to deny, or to make it configurable, if you prefer.

## How tested

`go test ./pkg/handler/...` with the new fixtures/tests in place but the handler unchanged (the reinvocation fixture `rawPodHasTokenVolume` passes both before and after, as intended):

```
--- FAIL: TestUpdatePodSpec (0.01s)
    --- FAIL: TestUpdatePodSpec/Pod_balajilovesoreos_in_file_testdata/rawPodHasForeignTokenVolume.pod.yaml (0.00s)
        handler_pod_test.go:173: Expected patch didn't match:
            Got
            	[{"op":"add","path":"/spec/containers","value":[{"name":"balajilovesoreos","image":"amazonlinux","env":[{"name":"AWS_ROLE_ARN",...},{"name":"AWS_WEB_IDENTITY_TOKEN_FILE","value":"/var/run/secrets/eks.amazonaws.com/serviceaccount/token"}],"resources":{},"volumeMounts":[{"name":"aws-iam-token","readOnly":true,"mountPath":"/var/run/secrets/eks.amazonaws.com/serviceaccount"}]}]}]
            Wanted:
            	null
    --- FAIL: TestUpdatePodSpec/Pod_balajilovesoreos_in_file_testdata/rawPodHasVolumeMount.pod.yaml (0.00s)
        handler_pod_test.go:173: Expected patch didn't match:
            Got
            	[{"op":"add","path":"/spec/containers","value":[{"name":"balajilovesoreos","image":"amazonlinux","env":[{"name":"AWS_ROLE_ARN",...},{"name":"AWS_WEB_IDENTITY_TOKEN_FILE","value":"/var/run/secrets/eks.amazonaws.com/serviceaccount/token"}],"resources":{},"volumeMounts":[{"name":"aws-iam-token","mountPath":""}]}]}]
            Wanted:
            	null
--- FAIL: TestMutatePod (0.00s)
    --- FAIL: TestMutatePod/ReservedVolumeNameUsedByAnotherVolume (0.00s)
        handler_test.go:94: Unexpected response. Got {"allowed": true, "patch": "W3sib3Ai...", "patchType": "JSONPatch"} wanted {"allowed": true}
FAIL	github.com/aws/amazon-eks-pod-identity-webhook/pkg/handler	0.228s
```

After the fix:

```
$ gofmt -l pkg/handler && go vet ./pkg/handler/...
$ go test -count=1 ./pkg/handler/...
W0908 00:10:14 handler.go:351] Pod was not mutated. Reason: volume "aws-iam-token" is reserved for the projected service account token, but the pod declares a HostPath volume with that name. Pod=balajilovesoreos, ServiceAccount=default, Namespace=default
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/handler	0.227s
$ go test -count=1 ./...
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg	0.006s
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache	0.121s
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/cache/debug	0.531s
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/cert	0.022s
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/containercredentials	2.009s
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/filesystem	3.013s
ok  	github.com/aws/amazon-eks-pod-identity-webhook/pkg/handler	0.230s
```

New coverage: fixtures `rawPodHasVolumeMount` (hostPath, no mutation), `rawPodHasForeignTokenVolume` (projected token for another audience, no mutation), `rawPodHasTokenVolume` (already injected; sidecar still gets env + mount, volume untouched), `containercredentials/rawPodWithConflictingVolumeMount` (no mutation); `TestMutatePod/ReservedVolumeNameUsedByAnotherVolume` asserts the admission response is `Allowed` with no patch; `TestFindTokenVolume` covers none / other name / hostPath / emptyDir / projected without token / other audience / other path / injected.

## Links

- Issue: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/315

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Self-review before marking ready (an AI agent I operate prepared this; I reviewed the diff and re-ran the tests). Two things worth calling out for maintainers: (1) the same name-only check exists on the EKS Pod Identity (container credentials) path, so the `rawPodWithConflictingVolumeMount` fixture expectation changed too; (2) skip-and-warn vs deny is your call, see the question in the description. Happy to adjust either.

